PR #1, opened 2015-05-20
Admiral AppBar

evanrs:
Depends on https://github.com/Thinkful/tfcommons/pull/129.

cc @ripleyaffect


src/AppBar/LinkSet.es6
```diff
@@ -0,0 +1,48 @@
+const config = {
+    officeHours: {
+        icon: 'users'
+    }
+}
+
+if (global.__env.config) {
+    let envConfig = global.__env.config;
+    Object.keys(envConfig).forEach(key => {
```
tholex: No multiline lambdas? We've discussed the inconsistent return behavior before.

src/Icon.jsx
```diff
@@ -0,0 +1,19 @@
...
+ */
+class Icon extends React.Component {
+    static propTypes = {
+        name: React.PropTypes.string.isRequired
+    }
+
+    render() {
+        return <span
+            aria-hidden="true"
+            className={`icon-${this.props.name} ${this.props.className || ''}`}/>
```
tholex: Does this replace the Component? Is there no method to "addClassName" or "prependClassName" on instantiation?

src/Icon.jsx
```diff
@@ -0,0 +1,19 @@
+const React = require('react');
+
+/**
+ * Icon
+ * @property {String} Icon the icon class to use
```
tholex: specify that it's the `name` prop?

src/Icon.jsx
```diff
@@ -0,0 +1,19 @@
...
+ */
+class Icon extends React.Component {
+    static propTypes = {
+        name: React.PropTypes.string.isRequired
```
tholex: @ripleyaffect :+1: on this ?

src/Gravatar.jsx
```diff
@@ -0,0 +1,29 @@
+const React = require('react');
+const MD5 = require('spark-md5');
+
+const URL = 'https://www.gravatar.com/avatar';
+
+/**
+ * Gravatar
+ * @property {string} email the users email to use with gravatar
```
tholex: looks like there's also style and size?

src/AppBar/app-bar.less
```diff
@@ -0,0 +1,245 @@
...
+  border-radius: 1px;
+
+  .navigation__visible & {
+    &::before {
+      content: "Close";
+    }
+  }
+}
+
+.profile-img-placeholder {
```
tholex: `app-nav-gravatar`?

src/Gravatar.jsx
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    static defaultProps = {
+        style: 'retro',
+        size: 200
+    }
+
+    render() {
+        return <img
+            className={`gravatar ${this.props.className || ''}`}
```
tholex: you can just remove `className` probably, as you specify the className elsewhere?

src/AppBar/Navigation.jsx
```diff
@@ -0,0 +1,95 @@
...
+        const navClassName = cx(
+            'app-nav', {navigation__visible: this.state.isMenuVisible});
+
+        const linkSet = new LinkSet(user);
+
+        return (
+          <div className="nav-container">
+                <nav className={navClassName} rel="main-navigation">
+                    <a className="app-nav-logo" href="/">
+                        <img src="//tf-assets-prod.s3.amazonaws.com/splash/logo/maze_icon_white_28px.svg" alt="Thinkful"/>
```
tholex: You can remove the `a` and make it a span w/o href instead.

Looks like you're not using the Masthead component, which would replace this. Inlined SVG gives us a lot of interested options later, so I'm :+1: on using Masthead if we can.

src/AppBar/Navigation.jsx
```diff
@@ -0,0 +1,95 @@
...
+    static propTypes = {
+        displayName: React.PropTypes.string,
+        icon: React.PropTypes.string,
+        url: React.PropTypes.string.isRequired
+    }
+
+    render() {
+        const {url, className, displayName, icon} = this.props;
+        return (
+            <a className={cx(className, "app-nav-link")} href={url}>
```
tholex: take the `cx` call into a let / const above? Crazy that react doesn't do this for you... :/ small footprint i guess.

src/AppBar/Navigation.jsx
```diff
@@ -0,0 +1,95 @@
...
+        return (
+          <div className="nav-container">
+                <nav className={navClassName} rel="main-navigation">
+                    <a className="app-nav-logo" href="/">
+                        <img src="//tf-assets-prod.s3.amazonaws.com/splash/logo/maze_icon_white_28px.svg" alt="Thinkful"/>
+                    </a>
+                    <ul className="app-nav-main">
+                        {linkSet.left.map(
+                            (link, key) => <li>
+                                <NavLink key={key} {...link}/></li>)}
```
tholex: Let's chat about this LinkSet stuff tomorrow, wanna know what's going on here...

src/AppBar/LinkSet.es6
evanrs: @tholex that's just you, I'm not changing this because you've got a disagreement with the ECMAScript spec.

src/AppBar/Navigation.jsx
evanrs: Not moving the cx call out of className assignment.

src/AppBar/Navigation.jsx
evanrs: Awesome, yea, fun loading animations and other fun geewiz effects.

src/AppBar/Navigation.jsx
evanrs: It's a [property spread in React](https://facebook.github.io/react/docs/jsx-spread.html).

evanrs: Examine the static properties in NavLink to know what's passed in
```es6
    static propTypes = {
        displayName: React.PropTypes.string,
        icon: React.PropTypes.string,
        url: React.PropTypes.string.isRequired
   }
```
if something is missing an error is thrown

src/Gravatar.jsx
evanrs: It's more about providing the right interface, if I want to override the style rules of the component it should be easy without wrapping it in another element.\

src/Icon.jsx
evanrs: As part of React?

src/AppBar/LinkSet.es6
evanrs: I don't think it's relevant here, forEach is non-returning to begin with.

src/AppBar/Navigation.jsx
evanrs: It's common to us cx within assignment to className, before React deprecated it's own tool some of their examples do it this way. When the operation is straight forward I see no problem with it.

src/AppBar/Navigation.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+        return (
+          <div className="nav-container">
+                <nav className={navClassName} rel="main-navigation">
+                    <Masthead className="app-nav-logo"/>
+                    <ul className="app-nav-main">
+                        {linkSet.left.map(
+                            (link, key) => <li>
+                                <NavLink key={key} {...link}/></li>)}
+                        {linkSet.right.map(
```
tholex: Don't need right-side links here, `app-nav-main` is just left. Sounds like we don't need `right` at all - just `main` (duplicated) and `menu` (all links not in main)

src/AppBar/Navigation.jsx
```diff
@@ -0,0 +1,92 @@
...
+                            (link, key) => <li>
+                                <NavLink key={key} {...link}/></li>)}
+                    </ul>
+                    <ul className="app-nav-list">
+                        {[].concat(linkSet.left, linkSet.right).map(
```
tholex: yup just left, which we're renaming to `main`.

src/AppBar/LinkSet.es6
```diff
@@ -0,0 +1,48 @@
...
+        let link = config[key] = config[key] || {};
+        let defaults = envConfig[key] || {};
+
+        Object.keys(defaults).forEach(
+            key => link[key] = link[key] || defaults[key])
+    })
+}
+
+class LinkSet {
```
tholex: class => `function`

src/AppBar/LinkSet.es6
```diff
@@ -0,0 +1,48 @@
+const config = {
+    officeHours: {
+        icon: 'users'
+    }
+}
+
+if (global.__env.config) {
+    let envConfig = global.__env.config;
+    Object.keys(envConfig).forEach(key => {
+        let link = config[key] = config[key] || {};
```
tholex: try to move/remove this to `config` assignment / simplify it, and just use `global.__env.config` in the linkset construction below

tholex:
Comments! otherwise :+1:

PR #23, opened 2015-05-22
support markdown in webpack

benjaminEwhite:


evanrs:
:+1: thanks @benjaminEwhite

PR #28, opened 2015-05-27
Standalone appbar

evanrs:


src/AppBar/linkSet.es6
```diff
@@ -1,4 +1,6 @@
-const _ = require('lodash');
+const assign = require('lodash/object/assign');
+const defaults = require('lodash/object/defaults');
+const mapValues = require('lodash/object/mapValues');
```
benjaminEwhite: Any reason not using `import` syntax here, since that seems to be the standard we're moving to?

src/AppBar/standalone.less
```diff
@@ -0,0 +1,8 @@
+@import '~tfstyleguide/iconfont';
+
+#TUI\.AppBar {
```
benjaminEwhite: What's up with the `\` in this selector?

benjaminEwhite:
Other than 2 small comments :+1:

src/AppBar/standalone.less
tholex: `.` would indicate a class... this escapes it meaning `TUI.AppBar` is the id. @evanrs is this a react convention?

src/AppBar/standalone.es6
```diff
@@ -0,0 +1,20 @@
+const React = require('react');
+
+require('./standalone.less')
+const AppBar = require('./Navigation');
+
+module.exports = {
+    AppBar,
+    mount() {
+        let mountElement = document.getElementById('TUI.AppBar');
```
tholex: `TUI_AppBar` maybe?

benjaminEwhite: Yeah, I like `TUI_AppBar` more. Feels weird to be mimicking dot notation in CSS-classes.

src/AppBar/linkSet.es6
evanrs: After a deeper examination of the transpiled code from Babel and some of the obnoxious difficulties it creates with unintuitive variable assignment, @ripleyaffect and I determined it's better to simplify here.

PR #42, opened 2015-06-01
Add shared dropdown element

learykara:
Adds shared dropdown element to thinkful ui.

### DIFF
--- a/src/Dropdown.jsx
+++ b/src/Dropdown.jsx
@@ -1,0 +1,93 @@
+const React = require('react');
+const classnames = require('classnames');
+
+require('./dropdown.less');
+
+
+/**
+ * Shared dropdown menu element
+ * @property data {Array} of items to display in the dropdown,
+      containing (optional) `slug` and `displayName`
+ * @property selectedInd {Int} of the selected index in the list
+ * @property defaultDisplay {String} of default value that should be displayed
+ * @property handleChange {Function} to handle dropdown click/change
+ */
+const Dropdown = React.createClass({
+
+  getInitialState() {
+    return {open: false};
+  },
+
+  getDefaultProps() {
+    return {data: []};
+  },
+
+  componentDidMount() {
+    document.addEventListener('click', (e) => this._checkClickAway(e));
+  },
+
+  componentWillUnmount() {
+    document.removeEventListener('click', (e) => this._checkClickAway(e));
+  },
+
+  _checkClickAway(e) {
+    if (!e.target.hasAttribute('data-clickable')) {
+      this.setState({open: false});
+    }
+  },
+
+  _generateNodes() {
+    let {data} = this.props;
+    return data.map((item, ind) => {
+      return (
+        <p
+          className="dropdown-item"
+          id={ind}
+          data-slug={item.slug}>
+          {item.displayName}
+        </p>
+      );
+    });
+  },
+
+  _toggleOpen() {
+    this.setState({open: !this.state.open});
+  },
+
+  _handleChange(event) {
+    let {handleChange} = this.props;
+    this._toggleOpen();
+    handleChange(event);
+  },
+
+  componentClickAway() {
+    this.setState({open: false});
+  },
+
+  render() {
+    const nodes = this._generateNodes();
+    let {selectedInd, data, defaultDisplay} = this.props;
+
+    let buttonClasses = classnames(
+      'button', 'button__white', 'dd-button');
+    let dropdownClasses = classnames(
+      'dd-open',
+      {'hidden': !this.state.open});
+
+    return (
+      <button
+        className={buttonClasses}
+        onClick={this._toggleOpen}
+        data-clickable>
+        {data[selectedInd] && data[selectedInd].displayName || defaultDisplay}
+        <div
+          className={dropdownClasses}
+          onClick={this._handleChange}>
+          {nodes}
+        </div>
+      </button>
+    );
+  }
+});
+
+module.exports = {Dropdown};
--- a/src/Dropdown.jsx
+++ b/src/Dropdown.jsx
@@ -7,12 +7,19 @@ require('./dropdown.less');
 /**
  * Shared dropdown menu element
  * @property data {Array} of items to display in the dropdown,
-      containing (optional) `slug` and `displayName`
+      containing `value` and `displayName`
  * @property selectedInd {Int} of the selected index in the list
  * @property defaultDisplay {String} of default value that should be displayed
  * @property handleChange {Function} to handle dropdown click/change
  */
 const Dropdown = React.createClass({
+
+  propTypes: {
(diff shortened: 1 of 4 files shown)


src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
+const React = require('react');
+const classnames = require('classnames');
+
+require('./dropdown.less');
+
+
+/**
+ * Shared dropdown menu element
+ * @property data {Array} of items to display in the dropdown,
```
ripleyaffect: I'd guess that each item needs something specific. Either break the item out into it's own component and document those props there or document the array objects here

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+ * @property defaultDisplay {String} of default value that should be displayed
+ * @property handleChange {Function} to handle dropdown click/change
+ */
+const Dropdown = React.createClass({
+
+  getInitialState() {
+    return {open: false};
+  },
+
+  getDefaultProps() {
```
ripleyaffect: `propTypes` is a great way to self-document and ensure the correct thing is being passed in as well

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+    return {data: []};
+  },
+
+  componentDidMount() {
+    document.addEventListener('click', (e) => this._checkClickAway(e));
```
ripleyaffect: I think `global` is preferable? @evanrs?

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+  },
+
+  componentWillUnmount() {
+    document.removeEventListener('click', (e) => this._checkClickAway(e));
+  },
+
+  _checkClickAway(e) {
+    if (!e.target.hasAttribute('data-clickable')) {
```
ripleyaffect: note to go though the way material-ui handles this later, since this seems a little hacky

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+      this.setState({open: false});
+    }
+  },
+
+  _generateNodes() {
+    let {data} = this.props;
```
ripleyaffect: const

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+    return data.map((item, ind) => {
+      return (
+        <p
+          className="dropdown-item"
+          id={ind}
+          data-slug={item.slug}>
```
ripleyaffect: camelCase

ripleyaffect: `data-slug` is too specific for this component following whatever convention is used in `select` elements (`value` I think)

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+  componentClickAway() {
+    this.setState({open: false});
+  },
+
+  render() {
+    const nodes = this._generateNodes();
+    let {selectedInd, data, defaultDisplay} = this.props;
+
+    let buttonClasses = classnames(
```
ripleyaffect: const. no need for linebreak here

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+      'button', 'button__white', 'dd-button');
+    let dropdownClasses = classnames(
```
ripleyaffect: const

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+      'dd-open',
+      {'hidden': !this.state.open});
+
+    return (
+      <button
+        className={buttonClasses}
```
ripleyaffect: second indent for visual separatio

src/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+        onClick={this._toggleOpen}
+        data-clickable>
+        {data[selectedInd] && data[selectedInd].displayName || defaultDisplay}
+        <div
+          className={dropdownClasses}
+          onClick={this._handleChange}>
+          {nodes}
```
ripleyaffect: you can just call `_generateNodes` here

webpack.hotload.js
```diff
@@ -6,6 +6,7 @@ var webpack = require('webpack');
 var WebpackDevServer = require('webpack-dev-server');
 
 module.exports = function (webpackConfig, options) {
+    options = options || {};
```
ripleyaffect: :+1:

ripleyaffect:
Looks pretty good to me. Let's go over the open questions together and with @evanrs so we can get this out :dancer:

src/Dropdown.jsx
learykara: array objects documented on L10

src/Dropdown.jsx
learykara: added

src/dropdown.less
```diff
@@ -0,0 +1,56 @@
+@import '~tfstyleguide/vars';
+
+.button.button__white.dd-button {
+  width: 420px;
```
evanrs: Seems like this should be a max-width of 320 for mobile. @tholex?

tholex: max-width: 100%; so it shrinks when the container gets smaller than the explicit `width`?

learykara: got it

src/Dropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+ * @property defaultDisplay {String} of default value that should be displayed
+ * @property handleChange {Function} to handle dropdown click/change
+ */
+const Dropdown = React.createClass({
+
+  propTypes: {
+    data: React.PropTypes.array.isRequired,
+    selectedInd: React.PropTypes.number,
+    defaultDisplay: React.PropTypes.string,
+    handleChange: React.PropTypes.func.isRequired
```
evanrs: This should be `onChange`, in general we should try to mimic conventions we find in material-ui.

https://github.com/callemall/material-ui/blob/master/src/radio-button.jsx#L100-L101

src/Dropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+      containing `value` and `displayName`
+ * @property selectedInd {Int} of the selected index in the list
+ * @property defaultDisplay {String} of default value that should be displayed
+ * @property handleChange {Function} to handle dropdown click/change
+ */
+const Dropdown = React.createClass({
+
+  propTypes: {
+    data: React.PropTypes.array.isRequired,
+    selectedInd: React.PropTypes.number,
```
evanrs: Why is this property name truncated?

src/Dropdown.jsx
evanrs: I'll submit a PR for this to do something like this:
https://github.com/callemall/material-ui/blob/master/src/mixins/click-awayable.js

PR #48, opened 2015-06-02
Hide dropdown.

### LINKED ISSUE #15: clicking on other part of screen right-hand nav dropdown should close the dropdown
right now you have to click "close"

tholex:
Fixes #15, could use a little CSS animation tuning (disappears instantly, not exactly sure why, but I may have been editing peacock's `nav.less` accidentally)


evanrs:
:+1:

PR #55, opened 2015-06-03
Uses name instead of menu.

evanrs:


src/AppBar/Navigation.jsx
```diff
@@ -60,6 +60,8 @@ class AppNav extends React.Component {
 
     render() {
         const {user, config} = this.props;
+        console.log(user);
```
tholex: log

diezguerra:
@evanrs is there a reason to hold this?

tholex:
This was held until we have a tooltip component or some way of showing that there's a menu there (on hover is fine)

evanrs:
Closing stale branch.

PR #61, opened 2015-06-04
Gravatar component accepts style prop.

evanrs:


src/Gravatar.jsx
```diff
@@ -10,20 +10,23 @@ const URL = 'https://www.gravatar.com/avatar';
 class Gravatar extends React.Component {
     static propTypes = {
         email: React.PropTypes.string.isRequired,
-        style: React.PropTypes.string,
-        size: React.PropTypes.number
+        gravatarStyle: React.PropTypes.string,
```
ripleyaffect: seems like this should be something like `default`. It can take a url as well as the fallback styles like "retro"

ripleyaffect:
:+1:

PR #74, opened 2015-06-10
Corrects React warnings, organizes for consistency

evanrs:


src/Sidebar/MenuItem.jsx
```diff
@@ -3,8 +3,6 @@ const classnames = require('classnames');
 
 const Icon = require('../Icon').Icon;
 
-require('./sidebarmenucomponents.less');
-
```
learykara: where do the styles for this come from now?

evanrs: It was being imported twice, unfortunately it may not be deduped and will actually include the styles twice. If there's a chance that `MenuItem` will be imported outside of the context of `SidebarMenu` then `MenuItem` specific styles should be split out. Ideally everything is as self contained as possible.

https://github.com/Thinkful/ui/pull/74/files#diff-4dac53637911d2eaaeb29a92770c7681R4

learykara: makes sense

learykara:
looks good :+1:

PR #81, opened 2015-06-18
Fixes and style changes when integrating into robin.

rkuykendall:
Small changes to make AvailabilityGrid actually importable, and some style changes to closer fit mockup.


evanrs:
:+1:

PR #83, opened 2015-06-19
fix dropdown styling

learykara:


src/DatePicker/DatePicker.es6
```diff
@@ -0,0 +1,133 @@
...
+  _generateDays() {
+    let {monthsNavigated, activeIndex} = this.state;
+
+    const startDay = moment().add(monthsNavigated, 'month').
+                            startOf('month').startOf('week').startOf('day');
+    const endDay = moment().add(monthsNavigated, 'month').
+                          endOf('month').endOf('week').startOf('day');
+    const totalDays = endDay.diff(startDay, 'days') + 1;
+
+    const days = [for (i of Array(totalDays).keys()) i].
```
evanrs: Clever!

evanrs:
Solid code @learykara definitely a :+1:

src/DatePicker/DatePicker.es6
diezguerra: what's the difference between `[for (i of Array(totalDays).keys()) i]` and ```Array(totalDays).keys()```?

learykara: `Array(totalDays).keys()` returns an `ArrayIterator`

diezguerra: I just figured out that totalDays is just an int, so you are just bending over backwards to build a range. Can you get lodash in scope so you can use `_.range`?

Something slightly less (but still) ugly:

```es6
const days = Array(totalDays).fill(0).map(
    (_, index) => {
        return {
            dateObj: dateObj: moment(startDay).add(index, 'day'),
            dayOfYear: moment(startDay).add(index, 'day').dayOfYear()
        });

diezguerra: Anyway, if not using the lodash solution, I'd add a comment explaining what the maneuver does and why.

learykara: will do

PR #96, opened 2015-06-24
New disabled state from properties

rkuykendall:


src/AvailabilityGrid/AvailabilityGrid.jsx
```diff
@@ -230,21 +231,23 @@ const AvailabilityGrid = React.createClass({
...
-        selectionMode: newMode,
-        selectionStartDay: startDay,
-        selectionStartSlot: startSlot
-      });
-    } else {
-      this.setState({
-        selectionMode: newMode
-      });
+    if (!this.props.disabled) {
+      if (newMode == 'selecting' || newMode == 'unselecting') {
```
diezguerra: I understand the code was there before, but shouldn't you be using triple =?

diezguerra:
lgtm :+1: (except for the css, that's not my thing).

src/AvailabilityGrid/AvailabilityGrid.jsx
rkuykendall: From what I can tell, it's exactly the same thing. I could change it though. @evanrs?

diezguerra: that's right, it should be the same thing. I think the JS default is to use the triple one by default

src/AvailabilityGrid/availability_grid.less
```diff
@@ -1,5 +1,26 @@
...
 
+.availability-grid.disabled {
+  color: @gray50;
+
+  .availability-grid-days {
+    color: @gray50;
+
+    .availability-grid-day {
+
+      .availability-grid-slot {
```
tholex: no need to nest these...

src/AvailabilityGrid/availability_grid.less
```diff
@@ -1,5 +1,26 @@
 @import "~tfstyleguide/vars";
 
+.availability-grid.disabled {
```
tholex: If you want to "perfectly mimic" the rest of our LESS, use `.availability-grid__disabled` as the modifier class, that way you can actually drop all of these into a block inside `.availability-grid` and still have it be searchable.

rkuykendall: Awesome, thanks for the help.

src/AvailabilityGrid/AvailabilityGrid.jsx
evanrs: @diezguerra is correct you should only ever use strict equals `===`

evanrs: But also favor regular expressions for simple string matches like this
```es6
if (/selecting/g.test(newMode) {
}
```

src/AvailabilityGrid/AvailabilityGrid.jsx
```diff
@@ -306,9 +309,14 @@ const AvailabilityGrid = React.createClass({
       );
     });
 
+    let classes = classNames({
```
evanrs:
```es6
classNames('availability-grid', {'availability-grid__disabled': this.props.disabled})
```

src/AvailabilityGrid/AvailabilityGrid.jsx
```diff
@@ -306,9 +309,14 @@ const AvailabilityGrid = React.createClass({
...
+      'availability-grid': true,
+      'availability-grid__disabled': this.props.disabled
+    });
+
     return (
       <div
-          className="availability-grid"
+          className={classes}
           onMouseEnter={this.handleMouseEnter}
```
evanrs: Inconsistent naming, either handleMouseEnter or onMouseEnter, @ripleyaffect preferes `_handleMouseEnter`, however on is fewer characters and just as transparent.

src/AvailabilityGrid/AvailabilityGrid.jsx
rkuykendall: Wait... what? Why would you do that? Regular expressions for simple string comparisons?

src/AvailabilityGrid/AvailabilityGrid.jsx
rkuykendall: Hm, I actually thought `handle` was what you called yours and `on` was what you passed down as a param. Not inconsistent, just clearer.

src/AvailabilityGrid/AvailabilityGrid.jsx
rkuykendall: Fixed on https://github.com/Thinkful/ui/pull/98

PR #100, opened 2015-06-25
send tfl students to library as home

learykara:


src/AppBar/linkSet.es6
```diff
@@ -39,6 +39,14 @@ else {
         if (/core/.test(user.student_type)) {
             defaults(home, config.courses);
         }
+        else if (/tfl/.test(user.student_type)) {
+            assign(home,{
```
tholex: `assign(home, {` sorry pedantic space :)

Also, because it's an else if, i think alums will still see courses as their home...

tholex:
:+1:

PR #102, opened 2015-06-26
use all props sent to gravatar

learykara:


learykara:
@bgribble yeah, that sounds right. good call, changing now

PR #111, opened 2015-07-05
adds authed vs. unauthed state to nav bar...

benjaminEwhite:
...determining menu choices, and whether or not icons displayed


src/AppBar/Navigation.jsx
```diff
@@ -18,16 +18,18 @@ class NavLink extends React.Component {
         active: React.PropTypes.bool,
         displayName: React.PropTypes.string,
         icon: React.PropTypes.string,
+        displayIcon: React.PropTypes.bool,
```
evanrs: Use CSS and not React. `.menu .icon { display: none }`

src/AppBar/Navigation.jsx
```diff
@@ -6,7 +6,7 @@ const uniqueId = require('lodash/utility/uniqueId');
 const {Icon} = require('../Icon');
 const {Gravatar} = require('../Gravatar');
 const {Masthead} = require('./Masthead');
-
+const {ThinkfulLogo} = require('./ThinkfulLogo');
```
evanrs: Don't prefix with Thinkful or Tf, we know who we are.

src/AppBar/Navigation.jsx
```diff
@@ -76,40 +78,77 @@ class AppNav extends React.Component {
...
 
-    render() {
-        const {user, config} = this.props;
+    renderAuthed(user, config) {
         const navClassName = cx(
             'app-nav', {'app-nav__visible': this.state.isMenuVisible});
 
         return (
-          <div className="app-nav-container">
+          <div className={cx('app-nav-container')}>
```
evanrs: Don't use cx here.

src/AppBar/Navigation.jsx
```diff
@@ -76,40 +78,77 @@ class AppNav extends React.Component {
...
                         <span alt="Menu" className="app-nav-burger"></span>
                         <Gravatar className="app-nav-gravatar" email={user.tf_login} size={120}/>
                     </a>
-                    }
                 </nav>
             </div>
         )
     }
+
+    renderUnauthed(config) {
```
evanrs: This should be a separate component, the mounting logic can take care of this.

src/AppBar/Navigation.jsx
```diff
@@ -76,40 +78,77 @@ class AppNav extends React.Component {
...
+                        <span alt='Menu' className='app-nav-burger'></span>
+                    </a>
+                </nav>
+            </div>
+        )
+    }
+
+    render() {
+        const {user, config} = this.props;
+        return _.has(user, 'tf_login') ?
```
evanrs: Re: separate component.

src/AppBar/ThinkfulLogo.jsx
```diff
@@ -0,0 +1,26 @@
+const React = require('react');
+
+const ThinkfulLogo = React.createClass({
```
evanrs: This would actually be the Masthead, and the T would be the logo. It's just historical that the other component kept its name.

src/AppBar/ThinkfulLogo.jsx
```diff
@@ -0,0 +1,26 @@
...
+    render() {
+        return (
+            <svg width="136px" height="28px" viewBox="0 0 136 28" version="1.1" xmlns="http://www.w3.org/2000/svg" {...this.props}>
```
evanrs: Lets take the time to import svg with webpack instead of wrapping them in components.

src/AppBar/linkSet.es6
```diff
@@ -17,13 +17,18 @@ config = mapValues(global.__env.config,
...
 if(! user) {
-    defaults(home, config.www);
+    menu.push(config.library);
+    menu.push(config.officeHours);
+    menu.push(config.forBeginners);
+    menu.push(config.pricing);
+    menu.push(config.signIn)
 }
 else {
+    main.push(home);
```
evanrs: :+1:

src/AppBar/linkSet.es6
```diff
@@ -17,13 +17,18 @@ config = mapValues(global.__env.config,
...
 let home = {displayName: 'Home', icon: 'home'};
-let main = [home]
+let main = [];
 let menu = [];
 
 if(! user) {
-    defaults(home, config.www);
+    menu.push(config.library);
+    menu.push(config.officeHours);
+    menu.push(config.forBeginners);
```
evanrs: Lets go with more specific titles for all splash/marketing resources, they'll change faster and be more numerous.

src/AppBar/linkSet.es6
```diff
@@ -17,13 +17,18 @@ config = mapValues(global.__env.config,
...
+    menu.push(config.pricing);
```
evanrs: Same comment.

src/AppBar/Navigation.jsx
```diff
@@ -76,40 +78,77 @@ class AppNav extends React.Component {
...
                 </nav>
             </div>
         )
     }
+
+    renderUnauthed(config) {
+        const navClassName = cx(
+            'app-nav', {'app-nav__visible': this.state.isMenuVisible});
+        return (
+          <div className={cx('app-nav-container', 'app-nav-container__unauthed')}>
```
evanrs: CX not needed here.

src/AppBar/Navigation.jsx
```diff
@@ -76,40 +78,77 @@ class AppNav extends React.Component {
...
-          <div className="app-nav-container">
+          <div className={cx('app-nav-container')}>
                 <nav onMouseLeave={this.handleMouseLeave.bind(this)}
                      className={navClassName} rel="main-navigation">
                     <Masthead className="app-nav-logo"/>
                     <ul className="app-nav-main">
                         {linkSet.main.map(
                             (link) => <li key={uniqueId(link)}>
-                                <NavLink {...link}/></li>)}
+                                <NavLink {...link} displayIcon={true} /></li>)}
```
evanrs: nix displayIcon prop, rely on css for showing/hiding.

src/AppBar/Navigation.jsx
benjaminEwhite: Are you saying to turn it into a pure render component? something like `UnauthedNav` and `AuthedNav`?

benjaminEwhite: Fine with doing that, but what's the advantage?

src/AppBar/ThinkfulLogo.jsx
benjaminEwhite: e.g., save it to a file and require it?

src/AppBar/linkSet.es6
benjaminEwhite: I think this is the correct level of specificity. It is also what is indeed in the config. "library", "officehours" and "for Beginners" is how we're internally referring to these sections of the site, and they also (happen to) correspond to the url structure.

src/AppBar/Navigation.jsx
benjaminEwhite: Done

src/AppBar/Navigation.jsx
benjaminEwhite: Done.

src/AppBar/ThinkfulLogo.jsx
benjaminEwhite: done.

src/AppBar/ThinkfulLogo.jsx
benjaminEwhite: fixed.

src/AppBar/linkSet.es6
evanrs: The first make sense as nouns, the last one doesn't, make them work as nouns.

src/AppBar/Navigation.jsx
benjaminEwhite: done.

src/AppBar/Navigation.jsx
benjaminEwhite: fixed

src/AppBar/linkSet.es6
benjaminEwhite: see above.

src/AppBar/Masthead.jsx
```diff
@@ -1,16 +1,18 @@
 const React = require('react');
 
-const Masthead = React.createClass({
```
evanrs: This component is no longer useful.

src/AppBar/Navigation.jsx
evanrs: They serve very different use cases. And it uses a react component as a giant if statement, that should happen before mounting, React.render(user? AppBar : NavBar)

src/AppBar/Navigation.jsx
```diff
@@ -23,14 +22,13 @@ class NavLink extends React.Component {
...
+
         return (
             <a className={cx({active}, className, "app-nav-link")}
                href={url}
                target={external ? "_blank" : "_self"}>
-                {icon &&
-                    <Icon className="app-nav-icon" name={icon}/>}
-                {displayName &&
-                    <span className="app-nav-text">{displayName}</span>}
+                {icon && <Icon className="app-nav-icon" name={icon}/>}
```
evanrs: The indentation and line break are for readability and splitting the logical short circuit from the rendered element.

package.json
```diff
@@ -57,6 +57,7 @@
     "eslint": "^0.22.1",
     "eslint-loader": "^0.12.0",
     "extract-text-webpack-plugin": "^0.8.1",
+    "file-loader": "^0.8.4",
```
evanrs: Check that this works as you expect, are these files coming out in the build?

src/AppBar/Navigation.jsx
benjaminEwhite: I can see that, but I disagree that they should be different components, because they share much of the same less, and all of the same component functions (e.g., `toggleMenu` etc.). Alternatives would be duplicating code (no!) or having a base class that both inherit from (which strikes me as overkill in this case).  Thoughts?

benjaminEwhite: @evanrs ^^

src/AppBar/Masthead.jsx
benjaminEwhite: retired.

src/AppBar/Navigation.jsx
benjaminEwhite: fixed.

src/AppBar/Navigation.jsx
evanrs: You can leave it as is, but it will be split into two components at a later date.

package.json
benjaminEwhite: as discussed in slack, ended up going with html-loader + dangerouslySetInnerHtml as solution.

evanrs:
:+1:

PR #119, opened 2015-07-13
Adds external stylesheet option.

evanrs:


evanrs:
Merging in advance of review to reduce build sizes in production.

PR #120, opened 2015-07-13
Fix logo position on mobile

evanrs:


evanrs:
Merging in advance of review to fix logo position.

PR #132, opened 2015-07-29
Update date picker and dropdown elements following work in roadrunner

learykara:
@evanrs  please take a look when you have a minute. Minor changes to style, default props, etc in date picker and dropdown

### DIFF
--- a/src/DatePicker/DatePicker.es6
+++ b/src/DatePicker/DatePicker.es6
@@ -77,11 +77,11 @@ class DatePicker extends React.Component {
                           endOf('month').endOf('week').startOf('day');
     const totalDays = endDay.diff(startDay, 'days') + 1;
 
-    const days = [for (i of Array(totalDays).keys()) i].
-                map(day => {
+
+    const days = _.map(Array(totalDays), (i, idx) => {
                   return {
-                    dateObj: moment(startDay).add(day, 'day'),
-                    dayOfYear: moment(startDay).add(day, 'day').dayOfYear()
+                    dateObj: moment(startDay).add(idx, 'day'),
+                    dayOfYear: moment(startDay).add(idx, 'day').dayOfYear()
                   }});
 
     activeIndex = (! activeIndex || activeIndex === -1) ?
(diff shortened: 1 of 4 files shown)


src/DatePicker/DatePicker.es6
```diff
@@ -80,12 +84,13 @@ class DatePicker extends React.Component {
                     dayOfYear: moment(startDay).add(day, 'day').dayOfYear()
                   }});
 
-    activeIndex = activeIndex ||
-      _.findIndex(days, {dayOfYear: moment().dayOfYear()});
+    activeIndex = (! activeIndex || activeIndex === -1) ?
+      _.findIndex(days, {dayOfYear: moment(defaultDate || '').dayOfYear()}) : activeIndex;
```
evanrs: The else clause of the ternary expression should sit on the next line when following a calculated value

src/Dropdown/Dropdown.jsx
```diff
@@ -16,6 +16,7 @@ const Dropdown = React.createClass({
 
   propTypes: {
     data: React.PropTypes.array.isRequired,
+    initialSelectedInd: React.PropTypes.number,
```
evanrs: Full words are preferable, perhaps `defaultIndex` and `activeIndex`?

src/Dropdown/Dropdown.jsx
```diff
@@ -16,6 +16,7 @@ const Dropdown = React.createClass({
...
     selectedInd: React.PropTypes.number,
     defaultDisplay: React.PropTypes.string,
     handleChange: React.PropTypes.func.isRequired
```
evanrs: The api for event handlers usually has `on` as prefix, so `onMouseIn` or `onChange`.

src/Dropdown/Dropdown.jsx
```diff
@@ -72,15 +74,19 @@ const Dropdown = React.createClass({
...
-    let {selectedInd, data, defaultDisplay} = this.props;
+    let {
+      initialSelectedInd, selectedInd, data,
+      defaultDisplay, className} = this.props;
 
-    const buttonClasses = classnames('button', 'button__white', 'dd-button');
-    const dropdownClasses = classnames(
+    selectedInd = selectedInd || initialSelectedInd;
+
+    const buttonClasses = cx('button', 'button__white', 'dd-button');
```
evanrs: `cx` isn't need here because it's no removing any falsey values or helping to concatenate variables.

evanrs:
Take a look at the comments, but otherwise solid :+1:

src/Dropdown/Dropdown.jsx
learykara: reluctant to change this now as there are now a few different apps using the dropdown and I'd have to go back and change individually in all of them, but will keep this in mind going forward

src/Dropdown/Dropdown.jsx
learykara: ditto above

PR #143, opened 2015-08-06
Only replace selectedInd if it's undefined (was replacing when 0 too

learykara:
This was replacing `selectedInd` when it was equal to `0`. Instead, check if it is `undefined` and then replace


evanrs:
:+1:

PR #148, opened 2015-08-11
Fix availability grid on Safari

rkuykendall:
- Don't use fill because Safari doesn't support it.


darrellsilver:
:+1:

darrellsilver:
closes https://github.com/Thinkful/Thinkful/issues/42

PR #154, opened 2015-08-13
add library to nav for mentors/admins/core students

learykara:


evanrs:
@tholex take note of the media query change.

:+1:

PR #160, opened 2015-08-20
Feature/demo page

tholex:
Adds a demo page showcasing the various components using https://github.com/rpominov/react-demo

@learykara @ripleyaffect check it! `script/bootstrap` and then hit `:5029`

### DIFF
--- a/client/DemoPage.jsx
+++ b/client/DemoPage.jsx
@@ -1,11 +1,73 @@
 const React = require('react');
+const Demo = require('react-demo');
+const moment = require('moment-timezone');
+
+const {
+  Gravatar,
+  Icon,
+  Dropdown,
+  DatePicker,
+  AvailabilityGrid
+} = require('../src')
 
 require('./styles/demo.less')
 
 class DemoPage extends React.Component {
   render() {
+    const sampleDropdownData = [
+      { value: 'pizza', displayName: "Mothefuckin' Pizza"},
+      { value: 'waffles', displayName: "Aw sick Waffles"},
+      { value: 'felafel', displayName: "I'm feelin Felafel"},
+      { value: 'bose', displayName: "No highs no lows, must be Bose"}
+    ];
+
     return (<div className="tui-demo-page">
-      <h1>Should look like h1</h1>
+      <h1>Thinkful UI</h1>
+      <h3>Gravatar</h3>
+      <Demo
+        target={Gravatar}
+        props={{
+          email: Demo.props.string('tholex@gmail.com'),
+          size: Demo.props.choices([60,120])
+        }} />
+
+      <h3>Icon</h3>
+      <Demo
+        target={Icon}
+        props={{
+          name: Demo.props.string('pointupright'),
+          className: Demo.props.string('additional-class')
+        }} />
+
+      <h3>Dropdown</h3>
+      <Demo
+        target={Dropdown}
+        props={{
+          data: Demo.props.constant(sampleDropdownData),
+          initialSelectedInd: Demo.props.choices([undefined,0,1,2]),
+          defaultDisplay: Demo.props.string("Choose something awesome"),
+          handleChange: Demo.props.callback.log(e => e.target.getAttribute('value'))
+        }} />
+
+      <h3>DatePicker</h3>
+      <Demo
+        target={DatePicker}
+        props={{
+          defaultDate: Demo.props.constant(moment()),
+          className: Demo.props.choices(['additional-class', '']),
+          handleChange: Demo.props.callback.log(date => moment(date).format('YYYY-MM-DD'))
+        }} />
+
+      <h3>AvailabilityGrid</h3>
+      <Demo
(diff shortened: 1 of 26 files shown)


client/DemoPage.jsx
```diff
@@ -0,0 +1,75 @@
...
+} = require('../src')
+
+require('./styles/demo.less')
+
+class DemoPage extends React.Component {
+  render() {
+    const sampleDropdownData = [
+      { value: 'pizza', displayName: "Mothefuckin' Pizza"},
+      { value: 'waffles', displayName: "Aw sick Waffles"},
+      { value: 'felafel', displayName: "I'm feelin Felafel"},
```
ripleyaffect: falafel*

client/DemoPage.jsx
```diff
@@ -0,0 +1,75 @@
...
+  DatePicker,
+  AvailabilityGrid
+} = require('../src')
+
+require('./styles/demo.less')
+
+class DemoPage extends React.Component {
+  render() {
+    const sampleDropdownData = [
+      { value: 'pizza', displayName: "Mothefuckin' Pizza"},
```
ripleyaffect: :smile: think of the children!

demo/__env.json
```diff
@@ -0,0 +1,135 @@
+{
```
ripleyaffect: this should be accessible via assets, no?

demo/views/index.ejs
```diff
@@ -0,0 +1,22 @@
...
+  <link rel="icon" href="//tf-assets-prod.s3.amazonaws.com/splash/logo/favicon.ico" type="image/x-icon">
+
+  <link type="text/css" rel="stylesheet" href="//fast.fonts.net/cssapi/30f402f3-e8e1-463e-ae1a-1987161cd5e8.css"/>
+  <!-- CSS is injected by webpack hotloader -->
+</head>
+<body>
+  <div id="tui-demo-app">
+  </div>
+
+  <script type="text/javascript">window.__env = <%- JSON.stringify(global_env) %></script>
```
ripleyaffect: such templating

src/Dropdown/dropdown.less
```diff
@@ -6,20 +6,24 @@
   position: relative;
 }
 
-.button.button__white.dd-button {
+.dropdown-text {
```
ripleyaffect: What are your opinions on the `style` prop for components? Any cases where it's preferable to using CSS?

src/Gravatar.jsx
```diff
@@ -8,6 +8,8 @@ const URL = 'https://www.gravatar.com/avatar';
  * @property {string} email the users email to use with gravatar
  */
 class Gravatar extends React.Component {
+    static displayName = "Gravatar"
```
ripleyaffect: :+1:!

src/Loader/Loader.jsx
```diff
@@ -0,0 +1,17 @@
+const cx = require('classnames');
+const React = require('react/addons');
+
+require('./loader.less');
+
+class Loader extends React.Component {
+  static propTypes = {
+    className: React.PropTypes.oneOfType([React.PropTypes.string, React.PropTypes.object])
```
ripleyaffect: Restricting to string type seems reasonable. allowing for objects introduces different behavior because elements don't support it by default, you can just expect people to call classnames on an object before they pass it in.

src/Loader/loader.less
```diff
@@ -0,0 +1,12 @@
+@import "~tfstyleguide/vars";
```
ripleyaffect: file necessary?

ripleyaffect:
lookin' good. few comments, mostly :+1:

diezguerra:
@tholex let me know if you want this deployed somewhere?

demo/__env.json
tholex: I _could_ have a link to `localhost:5007`, but since `thinkful-ui` is in the wild, didn't feel like adding that dependency

src/Dropdown/dropdown.less
tholex: There's going to be some cases where you need it - setting a dynamic background-image or some height/width measurement, but generally I don't think we should use `style` attributes at all. The CSS could be way better -- maybe all scoped under `tui`

src/Loader/Loader.jsx
tholex: ah that's true... :+1: -- the loader is still just a wip, I haven't built it since I couldn't see it anywhere - that's what got me to make the demo page :smiley:

tholex:
@diezguerra As cool as it is, I don't think we need to deploy it... maybe it could replace styleguide.thinkful.com by copy-pasting the templates from there (part of the reason I did this in ejs). The main benefit is just being able to see and play w/ the ui element without hacking node_modules in another app...

tholex:
I'll send another PR to fill in the Loader shortly, don't wanna pry it out of this PR.

PR #162, opened 2015-08-24
Modal window

tholex:
Simple modal that includes an X and a curtain to close it. Mounts in-place (so it has to be in a non-hidden, non-transformed div).

The demo includes a button in the center to close the modal but this is totally optional - the host application would chose what to put in the modal.

Should I add an optional width= prop?

screenshot 2015-08-24 13 36 16

### DIFF
--- a/client/HomePage.jsx
+++ b/client/HomePage.jsx
@@ -1,0 +1,22 @@
+const React = require('react');
+const {Link} = require('react-router');
+
+class HomePage extends React.Component {
+  render() {
+    return (<div className="tui-demo-page">
+      <h1>Thinkful UI</h1>
+      <p>Shared React components for various Thinkful apps. The nav above
+      doesn't respond to your actual login-state, but uses a fake <code>__env.js</code> inside
+      thinkful-ui.</p>
+      <h3><Link to="/demo">
+        Interactive Component Demos
+      </Link></h3>
+
+      <h3><Link to="/demo">
+        Try Modal
+      </Link></h3>
+    </div>);
+  }
+}
+
+module.exports = {HomePage};
--- a/client/HomePage.jsx
+++ b/client/HomePage.jsx
@@ -23,13 +23,21 @@ class HomePage extends React.Component {
       <p>Shared React components for various Thinkful apps. The nav above
       doesn't respond to your actual login-state, but uses a fake <code>__env.js</code> inside
       thinkful-ui.</p>
-      <h3><Link to="/demo">
-        Interactive Component Demos
-      </Link></h3>
-
-      <h3><a onClick={this.toggleModal.bind(this)}>
-        Open Modal
-      </a></h3>
+      <h3>
+        <Link to="/demo">
+          Interactive Component Demos
+        </Link>
+      </h3>
+      <h3>
+        <Link to="/demo/modal">
+          Demo Page w/ Modal
+        </Link>
+      </h3>
+      <h3>
+        <a onClick={this.toggleModal.bind(this)}>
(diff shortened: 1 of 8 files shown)


src/Modal/Modal.jsx
```diff
@@ -0,0 +1,59 @@
...
+      "tui-modal-wrapper__hidden": (!controlledByParent) && (!isOpen)
+    });
+    const modalClasses = cx("tui-modal-content", className);
+
+    return (
+      <div className={wrapperClasses}>
+        <div className="tui-modal-curtain" onClick={closeModal}/>
+        <div className={modalClasses}>
+          <a className="tui-modal-close-button" onClick={closeModal}>
+            <span aria-hidden="true" className="icon-close" />
```
learykara: why not use the tui `Icon` component?

src/Modal/Modal.jsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  render() {
+    const {className} = this.props;
+    const {controlledByParent, isOpen} = this.state;
+
+    const closeModal = this.props.close || this._closeModal.bind(this);
+
+    const wrapperClasses = cx("tui-modal-wrapper", {
+      "tui-modal-wrapper__hidden": (!controlledByParent) && (!isOpen)
```
learykara: so this only needs to be hidden if it's not controlled by the parent because if it is controlled in the parent (as in HomePage), then it only exists on the page when `modalIsOpen` is `true`, correct?

client/routes.jsx
```diff
@@ -22,7 +23,8 @@ class App extends React.Component {
 
 const routes = (
   <Route name="app" path="/" handler={App}>
-    <Route name="demo" path="demo" handler={DemoPage}/>
+    <Route name="demo" path="demo" handler={DemoPage}></Route>
```
learykara: why this change?

src/Modal/Modal.jsx
learykara: ah, just read the comments above. so yeah that makes sense

client/HomePage.jsx
```diff
@@ -0,0 +1,46 @@
...
+  render() {
+    const {modalIsOpen} = this.state;
+    return (<div className="tui-demo-page">
+      <h1>Thinkful UI</h1>
+      <p>Shared React components for various Thinkful apps. The nav above
+      doesn't respond to your actual login-state, but uses a fake <code>__env.js</code> inside
+      thinkful-ui.</p>
+      <h3><Link to="/demo">
+        Interactive Component Demos
+      </Link></h3>
```
learykara: I haven't seen this style yet with two tags opened/closed on the same lines.. is that preferable in some cases?

learykara:
looks good to me!

client/routes.jsx
tholex: I was about to make a subroute with a modal -- will add it shortly..

tholex:
@learykara updated w/ static modal on a  route

src/Modal/Modal.jsx
```diff
@@ -0,0 +1,60 @@
...
+ * to show it conditionally on a user action.
+ */
+class Modal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isOpen: true,
+      controlledByParent: !! props.close
+    }
+  }
```
ripleyaffect: here is where you bind functions. something like
```es6
this._myFunction = this._myFunction.bind(this);
```

ripleyaffect:
:+1:

PR #165, opened 2015-08-27
Feature/logged out nav 2.0

tholex:
screenshot 2015-08-27 12 02 10

### DIFF
--- a/src/AppBar/Navigation.jsx
+++ b/src/AppBar/Navigation.jsx
@@ -58,7 +58,9 @@ class AppNav extends React.Component {
         return (
             <div className='app-nav-container'>
                 <nav onMouseLeave={this._handleMouseLeave}
-                     className={navClassName} rel="main-navigation">
+                     className={navClassName}
+                     key="main-navigation"
+                     rel="main-navigation">
                     <a href={linkSet.home.url}><div dangerouslySetInnerHTML={{__html: require('./images/white_t_logo.svg')}}>
                     </div></a>
                     <ul className="app-nav-main">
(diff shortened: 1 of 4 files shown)


src/AppBar/Navigation.jsx
```diff
@@ -114,30 +100,102 @@ class AppNav extends React.Component {
...
+
+      return (
+        <div className="app-nav-course-dropdown">
+          <span className='app-nav-link'
+                onClick={this._toggleCourseDropdown}>Learn</span>
+          <div onMouseEnter={this._handleMouseEnter}
+               className={dropdownContentClasses}>
+            <div className="subheading">Engineer Workshops</div>
+              <div className="app-nav-courses">
+                <a className="app-nav-courses-link" href="//projects.thinkful.com/library/node.js/introduction/">
```
learykara: do we have access to `CONFIG` at all in here? would be nice if these pointed to `t.ful` on def

src/AppBar/Navigation.jsx
```diff
@@ -114,30 +100,102 @@ class AppNav extends React.Component {
...
+                  <img className="app-nav-courses-icon" src="//tf-assets-prod.s3.amazonaws.com/wow-next/course-icons/node.svg" />
+                  <span>Node.js Library</span>
+                </a>
+              </div>
+            <div className="subheading">1-on-1 Courses</div>
+              <div className="app-nav-courses">
+                <a className="app-nav-courses-link" href="//www.thinkful.com/courses/learn-ux-online">
```
learykara: same question about `CONFIG`

src/AppBar/Navigation.jsx
```diff
@@ -114,30 +100,102 @@ class AppNav extends React.Component {
...
+                  <img className="app-nav-courses-icon" src="//tf-assets-prod.s3.amazonaws.com/wow-next/course-icons/uxd.svg" />
+                  <span>User Experience Design</span>
```
learykara: why the `<span>`

src/AppBar/Navigation.jsx
```diff
@@ -114,30 +100,102 @@ class AppNav extends React.Component {
...
-                        {linkSet.main.map(
-                            (link) => <li key={uniqueId(link)}>
-                                <NavLink
-                                    className='app-nav-link__mobile-only'
-                                    {...link} /></li>)}
+                        {linkSet.insertCourseDropdown && this.renderCourseDropdown()}
+                        {linkSet.insertCourseDropdown && <li key="courseDropdown">
+                            <NavLink className='app-nav-link__mobile-only'
+                                     displayName='Courses'
+                                     url='//www.thinkful.com/courses' />
```
learykara: does this url exist?

src/AppBar/app-bar.less
```diff
@@ -251,6 +256,101 @@
...
+  }
+}
+
+/* Course dropdown content */
+.app-nav-courses {
+  margin-right: -15px;
+  margin-left: -15px;
+  margin-bottom: 15px;
+  white-space: normal;
+  .grid-container();
```
learykara: is there a difference between calling / not calling a mixin? I think I've been using this just as `.grid-container`

src/AppBar/app-bar.less
```diff
@@ -251,6 +256,101 @@
...
+.app-nav-courses-icon {
+  vertical-align: middle;
+  height: 30px;
+  width: auto;
+  margin-right: 15px;
+}
+
+
+
+/* Unauthed navigationg */
```
learykara: navigationg

learykara:
:+1:

src/AppBar/Navigation.jsx
tholex: It will look slightly different but this : https://wow.qaful.com/courses/ is what it's gonna be

src/AppBar/Navigation.jsx
tholex: React adds spans around plain text anyway, but in this case I actually have some CSS targeting it. Dunno, I guess I can remove.

src/AppBar/app-bar.less
tholex: no difference

src/AppBar/Navigation.jsx
tholex: I'll add an issue, it's def. something I was thinking about. We can add all the course links there too and generate this.

There's a dummy copy of a logged out __env.js in this repo, so the app links are available. I could just add a dummy logged in __env.json, but in the interest of time I'd prefer to keep the config as a placeholder: https://github.com/Thinkful/ui/blob/develop/demo/__env.json

Since this is open source - dunno if a local config would make sense.

PR #179, opened 2015-09-14
Make __env optional for linkSet

ripleyaffect:
Having a reference to `global.__env.user` in `linkSet` creates a dependency whenever you import something from the root of the `thinkful-ui`. This makes that value optional.


benjaminEwhite:
:+1:

PR #185, opened 2015-09-18
Use dev server certs for https

ripleyaffect:
[`webpack-dev-server` provides ssl certificates](https://github.com/webpack/webpack-dev-server/tree/master/ssl) to use when serving over HTTPS in development. This sets the config to use those certificates if the `https` option is set to a truthy value.


diezguerra:
:+1: 

Should this be in skeleton too?

PR #186, opened 2015-09-21
Feature/splash footer

learykara:
Adds a `Footer` component that matches the pelican footer to be used on logged-out version of React apps.

### DIFF
--- a/src/Footer/footer.less
+++ b/src/Footer/footer.less
@@ -7,6 +7,8 @@
   .button {
     text-transform: uppercase;
     text-align: center;
+    margin-left: auto;
+    margin-right: auto;
 
     @media @nonmobile {
       display: none;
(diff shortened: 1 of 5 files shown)


src/Footer/footer.less
```diff
@@ -0,0 +1,137 @@
...
+  @media @nonmobile {
+    flex-basis: (@column-px * 3) + (@gutter-px * 2);
+  }
+
+  .footer-main-link {
+    color: @gray50;
+    display: block;
+    @media @nonmobile {
+      color: @white;
+      .body-text-small;
```
tholex: contact-links need a hover on desktop (here)

src/Footer/footer.less
```diff
@@ -0,0 +1,137 @@
...
+  margin-right: auto;
+  padding-top: 40px;
+  padding-bottom: 40px;
+  text-align: center;
+  display: flex;
+  flex-direction: column;
+
+  @media @nonmobile {
+    .centered-container;
+    .grid-container;
```
tholex: if you're using `display: flex` here you don't need `grid-container`

src/Footer/footer.less
```diff
@@ -0,0 +1,137 @@
...
+    text-align: left;
+    flex-direction: row;
+    justify-content: space-between;
+    padding-top: 80px;
+    padding-bottom: 80px;
```
tholex: multiple of 15px -- 75px?

tholex:
Watch out for http://caniuse.com/#feat=flexbox -- ~35% of IE users and ~70% of android users don't have this.

:+1: but only cause we need to move fast.

src/Footer/footer.less
```diff
@@ -0,0 +1,137 @@
...
+  .body-text-small;
+  display: none;
+  color: @blue75;
+  margin-left: 8px;
+  @media @nonmobile {
+    display: inline-block;
+  }
+}
+
+.social-links {
```
tholex: `footer-social-links`? seems too generic.

tholex:
If you're relying on flexbox this much, might as well get rid of `social-links__mobile` and __desktop. If the parent element is `display: flex;` you can use `order` to move the link to the top position on mobile instead of making two options. https://developer.mozilla.org/en-US/docs/Web/CSS/order

tholex:
`footer-social-links` and get rid of the `__mobile / __desktop` elements and it's :+1: 

Don't make more public-facing `display: flex;` layouts though, mentor-only pages.

src/Footer/footer.less
learykara: copied what is on pelican currently, but I will bump it down to 75

learykara:
closing to reopen against master

PR #192, opened 2015-09-29
Only use history if history exists

diezguerra:
Otherwise send folks to projects.

https://github.com/Thinkful/Thinkful/issues/126


diezguerra:
@tholex review sil vous plait?

tholex:
:+1:

PR #194, opened 2015-10-01
Analytics improvements for future use.

rkuykendall:


diezguerra:
@rkuykendall what up with this?

rkuykendall:
@diezguerra Wrote it but didn't have time to actually test and deploy. It's my pattern for how we should be doing analytics in the future. I think this is still a slower week on team withdraws so I'll try to verify and release it.

diezguerra:
Cool. Grab a copy of that product workflow book (there's many in the office), it should give you a good idea of why letting things sit is bad.

src/analytics/TrackedLink.jsx
```diff
@@ -35,8 +35,11 @@ class TrackedLink extends React.Component {
 
     _handleClick(event) {
         let data = omit(this.props, 'children', 'className', 'href', 'target');
+        data = {url: this.props.href, ...data};
+        let event = `clicked-${global.__env.config.appDisplayName}-${this.props.type}`;
```
learykara: `config` should come through props-- avoid using `global` wherever possible

src/analytics/TrackedLink.jsx
```diff
@@ -35,8 +35,11 @@ class TrackedLink extends React.Component {
 
     _handleClick(event) {
         let data = omit(this.props, 'children', 'className', 'href', 'target');
+        data = {url: this.props.href, ...data};
```
learykara: hmm - you're omitting `href` from `this.props` in creating `data`, but then adding it back in with the name `url`?

src/analytics/TrackedLink.jsx
learykara: also any reason `event` shouldn't be a `const`?

src/analytics/TrackedLink.jsx
```diff
@@ -35,8 +35,11 @@ class TrackedLink extends React.Component {
...
+        let event = `clicked-${global.__env.config.appDisplayName}-${this.props.type}`;
+        log(event, data);
```
learykara: is this `log` in here intentional?

src/analytics/actions.es6
```diff
@@ -1,20 +1,57 @@
...
+    // Temporary stub, which can be expanded upon to add data.
+    global.analytics &&
+        global.analytics.page.apply(this, arguments);
+}
+
 module.exports = {
-    track
+    track: track,
+    identify: identify,
+    page: page
```
learykara: can this just be
```
module.exports = {
    track,
    identify,
    page
}
```
?

src/analytics/actions.es6
```diff
@@ -1,20 +1,57 @@
...
+// This event mirrors the call signature of global.analytics.track
+function track(event, properties, options, fn) {
+    // Argument reshuffling, from original library.
+    if (is.fn(options)) fn = options, options = null;
+    if (is.fn(properties)) fn = properties, options = null, properties = null;
+
+    appInfo = {
         app: result(global.__env.config, 'app.name', '').toLowerCase(),
         appDisplayName: result(global.__env.config, 'app.displayName', '').toLowerCase(),
         ...data
```
learykara: where does `data` come from in this fn?

src/analytics/actions.es6
```diff
@@ -1,20 +1,57 @@
...
     }
+    let __env = __env || {};
```
learykara: does this pick up `global.__env`?

src/analytics/TrackedLink.jsx
learykara: oh wait, it's an arg. this throws an error

src/analytics/actions.es6
```diff
@@ -1,20 +1,57 @@
 const log = require('debug')('ui:analytics');
 const result = require('lodash/object/result');
+const Qs = require('qs');
 
-function track(action, type, data={}) {
-    action = `${action}: ${global.__env.config.appDisplayName}-${type}`;
-    data = {
+function mergeIntoDict(dest, src) {
+    let src = src || {};
+    let dest = dest || {};
```
learykara: `src` and `dest` are args - can't be redefined here

src/analytics/actions.es6
```diff
@@ -1,20 +1,57 @@
 const log = require('debug')('ui:analytics');
 const result = require('lodash/object/result');
+const Qs = require('qs');
```
learykara: does this need to be added to package.json?

src/analytics/actions.es6
```diff
@@ -1,20 +1,57 @@
...
+        }
+    }
+
+    return dest;
+}
+
+// This event mirrors the call signature of global.analytics.track
+function track(event, properties, options, fn) {
+    // Argument reshuffling, from original library.
+    if (is.fn(options)) fn = options, options = null;
```
learykara: what is `is.fn`? it's throwing an error for me locally

src/analytics/actions.es6
```diff
@@ -1,20 +1,57 @@
...
+const Qs = require('qs');
 
-function track(action, type, data={}) {
-    action = `${action}: ${global.__env.config.appDisplayName}-${type}`;
-    data = {
+function mergeIntoDict(dest, src) {
+    let src = src || {};
+    let dest = dest || {};
+
+    for (var key in src) {
```
learykara: `let key in src`

learykara:
I'm getting a lot of errors trying to use this locally and no data flowing through to segment.  Try to get it up and running locally and then we can get this out

src/analytics/TrackedLink.jsx
rkuykendall: Yeah, renaming it to URL. Would there be a better way? I actually didn't write the code just moved it into this function, but it seems to make sense.

learykara: no I can't come up with a better way either, this is fine

src/analytics/TrackedLink.jsx
```diff
@@ -35,8 +36,11 @@ class TrackedLink extends React.Component {
 
     _handleClick(event) {
         let data = omit(this.props, 'children', 'className', 'href', 'target');
+        data = {url: this.props.href, ...data};
+        let eventName = `clicked-${global.__env.config.appDisplayName}-${this.props.type}`;
```
learykara: config should come from props

src/analytics/actions.es6
```diff
@@ -1,20 +1,58 @@
...
+function track(event, properties, options, fn) {
+    // Argument reshuffling, from original library.
+    if (is.fn(options)) fn = options, options = null;
+    if (is.fn(properties)) fn = properties, options = null, properties = null;
+
+    const __env = global.__env || {};
+
+    let appInfo = {
+        app: result(__env.config, 'app.name', '').toLowerCase(),
+        appDisplayName: result(__env.config, 'app.displayName', '').toLowerCase()
```
learykara: config should come from props

src/analytics/actions.es6
```diff
@@ -1,20 +1,58 @@
...
+
+    global.analytics &&
+        global.analytics.track(event, properties, options, fn);
+}
+
+// This event mirrors the call signature of global.analytics.identify
+function identify(id, traits, options, fn) {
+    // Temporary stub, which can be expanded upon to add data.
+    global.analytics &&
+        global.analytics.track.apply(this, arguments);
```
learykara: where does `arguments` come from?

src/analytics/actions.es6
```diff
@@ -1,20 +1,58 @@
...
+}
 
+// This event mirrors the call signature of global.analytics.page
+function page(category, name, properties, options, fn) {
+    // Temporary stub, which can be expanded upon to add data.
     global.analytics &&
-        global.analytics.track(action, data);
+        global.analytics.page.apply(this, arguments);
```
learykara: same question about `arugments`

learykara:
a couple questions, but mostly looks good! :+1:

src/analytics/actions.es6
rkuykendall: In JS, arguments is all the arguments passed into a function.

Yeah, it didn't make sense to me either, but it's true. Yay magic variable!

src/analytics/actions.es6
rkuykendall: Below

src/analytics/actions.es6
rkuykendall: Hm, would you mind elaborating? This is a set of functions, not a react component, so I don't think it has props? These wrappers are designed to let us call things like `analytics.track('clicked-big-button')` and then be able to populate that track event with extra data like the app name and user information, etc.

learykara: you're right - ignore me - been a long week :grin:

PR #196, opened 2015-10-08
Add copy to the empty notification list

tholex:


src/NotificationView/NotificationView.jsx
```diff
@@ -51,6 +64,8 @@ class NotificationView extends React.Component {
       {"tui-notification-count__clear" : unseenCount === 0 }
     )
 
+    const hasNotifications = !_.isEmpty(notifications);
```
diezguerra: If `notifications` is not Empty but all notifications are read (which is a common case if you've x'd out all of yours), `hasNotifications` will be true but `renderNotifications` won't render any cause the `map` clause will filter all of them out. Shouldn't we do a filter here?

```es6
const hasNotifications = ! _.isEmpty(notifications.filter(notif => ! notif.is_read));
```

diezguerra:
:+1: except for the comment

PR #201, opened 2015-10-22
Feature/open sessions shared

learykara:


src/OpenSessionOverview/open-session-overview.less
```diff
@@ -0,0 +1,252 @@
...
+    }
+    .render-mobile & {
+      padding-top: 0;
+      max-width: 80%;
+    }
+  }
+}
+
+.session-tags {
+  display: flex;
```
tholex: Seems like untouched display: block works just as well but ok :)

tholex:
Stepped through the logic, makes sense. :+1:

tholex:
Since the whole component is built via props, consider dropping it in a `Demo` component on the demo page

learykara:
ah,  i always forget about demo. will do that

src/OpenSessionOverview/open-session-overview.less
learykara: was left over from when I copy-pasted - will update

PR #202, opened 2015-10-22
Add page, load, and identify commands. Fallback to Oilbird if analytics.js isn't working.

rkuykendall:


src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
+    if (urlParams.email) {
+        return decodeURIComponent(urlParams.email).toLowerCase()
+    }
 
-    let appInfo = {
-        app: result(__env.config, 'app.name', '').toLowerCase(),
-        appDisplayName: result(__env.config, 'app.displayName', '').toLowerCase()
+    // Check the form fields
+    let emailFields = document.querySelectorAll('[name="email"]');
+    if (emailFields.length && emailFields[0].value.length) {
```
learykara: is this guaranteed to have a `.value` ? `emailFields.length && emailFields[0].value && emailFields[0].value.length` would be _extra_ careful

src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
-    properties = mergeIntoDict(properties, __env.user);
-    properties = mergeIntoDict(properties, Qs.parse(window.location.search));
+// Failsafe for if segment breaks for some reason
+function fallback(callback, postData) {
+    superagent.
+        post(`${__env.config.oilbird.url}/echo`).
+        send(postData).
+        withCredentials().
+        end((error, response) => {
+            typeof callback === 'function' && callback();
```
learykara: notice you're using `is` elsewhere for this kind of thing - any reason not to use it here?

src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
+        });
+}
 
-    global.analytics &&
-        global.analytics.track(event, properties, options, fn);
+function mountSegmentIO() {
+    let analytics = global.analytics = global.analytics || [];
```
learykara: what does this `global.analytics` on both sides of the `=` do?

src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
+    if (!analytics.initialize) {
+        if (analytics.invoked) global.console && console.error && console.error("Segment snippet included twice.");
+        else {
+            analytics.invoked = !0;
```
learykara: huh? why not just `true`?

src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
+    // Check the URL parameters
+    if (urlParams.email) {
+        return decodeURIComponent(urlParams.email).toLowerCase()
+    }
 
-    let appInfo = {
-        app: result(__env.config, 'app.name', '').toLowerCase(),
-        appDisplayName: result(__env.config, 'app.displayName', '').toLowerCase()
+    // Check the form fields
+    let emailFields = document.querySelectorAll('[name="email"]');
```
learykara: `const`?

src/analytics/actions.es6
learykara: also, `const`?

src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
-    global.analytics &&
-        global.analytics.track(event, properties, options, fn);
+function mountSegmentIO() {
+    let analytics = global.analytics = global.analytics || [];
+    if (!analytics.initialize) {
+        if (analytics.invoked) global.console && console.error && console.error("Segment snippet included twice.");
+        else {
+            analytics.invoked = !0;
+            analytics.methods = ["trackSubmit", "trackClick", "trackLink", "trackForm", "pageview", "identify", "group", "track", "ready", "alias", "page", "once", "off", "on"];
+            analytics.factory = function(t) {
```
learykara: what is `t` / where does it come from?

src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
+    if (is.object(id)) options = traits, traits = id, id = user.id();
+
+    let email = tryEmail();
+    if (email && !id) {
+        id = email;
+
+        // Check if we previously had a different email on file
+        // for the user, and alias this email address to it
+        if (cookies.user_email && (email !== cookies.user_email)
+                && (cookies.user_email.indexOf('@thinkful.com') > -1)) {
```
learykara: don't you want the opposite? only alias if `@thinkful.com` index is -1?

src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
+            alias(email, alias.user_email);
```
learykara: `cookies.user_email`?

src/analytics/actions.es6
```diff
@@ -16,43 +28,184 @@ function mergeIntoDict(dest, src) {
...
+}
+
+// This event mirrors the call signature of global.analytics.alias
+function alias(to, from, options, fn) {
+    // Argument reshuffling, from original library.
+    if (is.fn(options)) fn = options, options = null;
+    if (is.fn(from)) fn = from, options = null, from = null;
+    if (is.object(from)) options = from, from = null;
+
+    // TODO: Don't alias thinkful.com emails.
```
learykara: I think you can remove this `TODO` based on L143 right?

learykara:
some questions, otherwise :+1:

src/analytics/actions.es6
rkuykendall: Double checked it just to be safe. As long as a field exists, it will have a value, and we know it exists due to `emailFields.length`. An empty field has the value `""`. and a `value.length` of 0.

src/analytics/actions.es6
rkuykendall: No this is a really great place for it, thanks! I was really only importing `is` so I could make sure I was duplicating the Segment argument reordering perfectly, but if I'm importing it I should use it for cases like this where it makes sense.

src/analytics/actions.es6
rkuykendall: So I have no idea, but I'm kinda inclined not to mess with it much since this code is directly from Segment, as the code you're supposed to include in your thing:

https://segment.com/docs/libraries/analytics.js/quickstart/

So I actually don't understand most of it, but it's what they supply for adding segment to your app. In fact, I think analytics is supposed to not be const, since this is a dummy analytics that gets overwritten by the real analytics script which you can see it adding to the source below.

src/analytics/actions.es6
rkuykendall: See above rationale for leaving this unaltered.

src/analytics/actions.es6

src/analytics/actions.es6
rkuykendall: Commenting on third alias note.

src/analytics/actions.es6
rkuykendall: Commenting on third alias note.

src/analytics/actions.es6
rkuykendall: Technically it wasn't being done in alias where it should have been, but in that one function. I'm actually on the fence about which it should be in but I moved it into alias for now. And yeah, the stuff you commented on above was wrong, sorry about that and thanks for catching it.

src/analytics/actions.es6
learykara: ah, I wasn't sure if it was directly from segment or not - I'll assume it's fine that I don't understand then :smile:

src/analytics/actions.es6
learykara: yep, makes sense. wasn't sure if there was some javascript sorcery I was missing out on

PR #205, opened 2015-10-26
add underlay and click away event

learykara:
Adds an underlay to the search bar that fades from white to transparent. Clicking away from the search bar closes search

image


rkuykendall:
:+1: To everything, but maybe get Olex to look over the CSS because that's some black magic. Although I guess if it works it's all good, it sure reads well.

PR #206, opened 2015-10-26
make nav and search bar work on mobile

learykara:


src/AppBar/NavLink.jsx
```diff
@@ -50,9 +50,14 @@ class SearchLink extends React.Component {
         this.state = {open: false}
     }
 
-    _toggleSearchForm = event => {
+    _handleSearchClick = event => {
```
diezguerra: is using the fat arrow a way to avoid having to bind this one in the constructor?

learykara: yep. thomas discovered that trick



—

benjaminEwhite: glad to learn this one amazing trick.

src/AppBar/NavLink.jsx
```diff
@@ -50,9 +50,14 @@ class SearchLink extends React.Component {
...
         event.preventDefault();
-        this.setState({open: !this.state.open});
+        const {active, config, mobile} = this.props;
```
benjaminEwhite: probably not a big deal, but `active` is a little vague, doesn't tell me what is active.

benjaminEwhite: would `isMobile` / `isActive` work here? haven't dug enough to see if that's what these vars are really about.

benjaminEwhite:
:+1:

src/AppBar/NavLink.jsx
learykara: this is the existing convention used in the other nav links - I agree it could be more descriptive, but want to avoid changing the other components for now

src/AppBar/NavLink.jsx
diezguerra: The closure trick the react core committers don't want you to know about.
React core devs hate this time-saving closure trick.
This FE developer was so touched he started crying, here's why.

learykara: I would _totally_ read that article #clickbait

PR #207, opened 2015-10-27
Add pathWithSlug/pathWithoutSlug utils for reverse proxied apps

bgribble:
The slug is config.app.slug from the JavaScript __env.  For reverse-proxied apps formerly known as
`appname.thinkful.com` this will be `/appname`


tati:
Explained IRL by Bill! :+1:

PR #208, opened 2015-10-28
Update Course Title

tati:


learykara:
don't forget to bump the version number :+1:

learykara:
actually, I'm going to release something later today if you want to send both of those out under one version

tati:
@learykara that would be awesome...gonna look at this locally one time and make sure it doesn't go to next line for justin real quick but should be good to go

tati:
Jk justin needs this out asap because now the course page doesn't match the menu

PR #214, opened 2015-11-03
Show dashboard link for students with core-student access flag.

rkuykendall:


bgribble:
:+1: delicious, locally-sourced, and completely GMO-free!

PR #217, opened 2015-11-05
Feature/search bar improvements

learykara:


src/AppBar/NavLink.jsx
```diff
@@ -51,6 +54,25 @@ class SearchLink extends React.Component {
...
+    componentDidMount() {
+        window.addEventListener('keydown', this._handleWindowKeyDown);
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener('keydown', this._handleWindowKeyDown);
+    }
+
+    _handleWindowKeyDown = (event) => {
+        if (event.which === SLASH_KEY_CODE) {
```
ripleyaffect: will this make the search bar pop up if typing in the raven editor?

src/AppBar/NavLink.jsx
```diff
@@ -51,6 +54,25 @@ class SearchLink extends React.Component {
...
+            this.setState({open: true});
+            this.refs.searchBar._autoFocus();
```
ripleyaffect: you shouldn't be calling an "internal" method like this, from outside the component. you can expose it with a public facing method called `autoFocus` which just calls the internal one or you can remove the underscore at the beginning

src/AppBar/NavLink.jsx
```diff
@@ -51,6 +54,25 @@ class SearchLink extends React.Component {
...
+        }
+        else if (event.which === ESC_KEY_CODE) {
+            this.setState({open: false});
+            this.refs.searchBar._unFocus();
```
ripleyaffect: ditto on the api

src/AppBar/NavLink.jsx
```diff
@@ -66,7 +88,7 @@ class SearchLink extends React.Component {
 
         mobile ?
             window.location = `${config.projects.url}/search`
-        :   active ? this.refs.searchBar._autoFocus()
+        :   active ? this.refs.searchBar._wiggle()
```
ripleyaffect: tritto on the api

src/SearchBar/SearchBar.jsx
```diff
@@ -54,7 +54,21 @@ class SearchBar extends React.Component {
...
+    React.findDOMNode(this.refs.input).value = this.state.searchTerm;
+  }
+
+  _unFocus = () => {
+    React.findDOMNode(this.refs.input).blur();
+  }
+
+  _wiggle = () => {
+    this._autoFocus();
+    React.findDOMNode(this.refs.input).classList.add('wiggle');
```
ripleyaffect: You shouldn't be adding classes like this in React, you can call an exposed api method on the component, something like "addClass", which would add it to the component's state, and then rerender it with the class added to `className`

src/SearchBar/SearchBar.jsx
```diff
@@ -54,7 +54,21 @@ class SearchBar extends React.Component {
...
+    window.setTimeout(
+      () => React.findDOMNode(this.refs.input).classList.remove('wiggle'), 1000);
```
ripleyaffect: ditto on adding/removing clases

ripleyaffect:
Address comments then :+1:

src/AppBar/NavLink.jsx
learykara: yes, as a matter of fact, it will. damnit.

PR #220, opened 2015-11-06
Feature/text loader

benjaminEwhite:


tati:
:+1: I don't totally understand but syntax look fine to me!

PR #221, opened 2015-11-06
Feature/high z index modal

benjaminEwhite:
Otherwise stuff below the modal appears above it :(

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "thinkful-ui",
-  "version": "1.1.48",
+  "version": "1.1.49",
   "description": "Shared navigation and UI resources for Thinkful.",
   "main": "src/index.es6",
   "scripts": {
--- a/src/Modal/modal.less
+++ b/src/Modal/modal.less
@@ -1,7 +1,7 @@
 @import "~tfstyleguide/vars";
 
 .tui-modal-wrapper {
-  z-index: 100;
+  z-index: 15;
   position: fixed;
   left: 0;
   right: 0;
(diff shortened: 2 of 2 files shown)


learykara:
`100` seems a little high - is it really necessary to be that large?

benjaminEwhite:
@learykara it's a guarantee that this will be on top. Think that's not bad practice, but correct me if i'm wrong.

benjaminEwhite:
And i think, by definition, a modal should always be on top, no?

learykara:
yeah, but I think it's best practice to make it as high as it needs to be without being excessively high. I remember being scolded once upon a time (I think by evan) for having unnecessarily high z-indexes -- @tholex care to weigh in?

benjaminEwhite:
My thinking is that this is in Thinkful ui, so we don't really know what z-indexes will be in apps that use it. Setting it at 100 is a guarantee that it will always be higher than what is in any specific app. 

I mean, what constitutes high enough? 50? 25?

learykara:
I believe the nav bar, which is on top of anything except a modal, is set at 10 -- but I'd have to double check

benjaminEwhite:
@learykara that is correct. what do you think, should a modal appear above or below navbar?

benjaminEwhite:
Vibrant discussion on z-index!

learykara:
modal should be above -- so a z-index of 15 should be plenty ?

benjaminEwhite:
You've got a deal!

learykara:
:+1:

tholex:
I thought the modal already had ~100 zindex. Sorry I didn't comment.

tholex:
We can put some z-indexes into vars.less in the styleguide – variables like `z-modal`, `z-nav`, and if you want to put something above the modal you just say `z-index: @z-modal + 1;`

PR #223, opened 2015-11-06
Update babel-core

benjaminEwhite:
had a bug with using `encodeURIComponent()` inside es6 template strings, and upgrading babel-core was the fix.

The '/' in `${encodeURIComponent('https://something.com')}` were incorrectly coded.


tholex:
:+1:

PR #228, opened 2015-11-11
Feature/update homes

ripleyaffect:


src/AppBar/linkSet.es6
```diff
@@ -45,39 +45,42 @@ if(! user) {
...
+            main.push(home);
+            main.push(config.paths);
         }
         else if (/tfl/.test(user.student_type)) {
-            assign(home, {
-                displayName: 'Library',
-                icon: 'book',
-                host: config.dashboard.host,
-                url: `${config.dashboard.url}/library`
+            home = assign(home, {
```
learykara: can you help me explain the difference between `home = assign(home, {...})` and `defaults(home, config...)`?

ripleyaffect: `assign` overwrites values in the first object with values on the later objects. `defaults` falls back to the later values if they aren't present in the first object, but will not overwrite existing values. Both do in-place mutation on the first object though, so this reassignment isn't really doing anything. For redux, I've been creating new objects using `newState = _.assign({}, oldState, newValues)` because you need new objects every time for some of the features. I'll remove the assignment here.

learykara:
:+1:

PR #230, opened 2015-11-12
Update to React 0.14

ripleyaffect:
- Upgrades `react`, `react-router`, and `react-demo`
- Adds `react-dom` library
- Makes changes required for update, [specified here](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html)


tholex:
check `NavLink`  - it calls react component methods on `refs.searchBar` which, so i assume that functionality will break...

tholex:
:+1:

PR #232, opened 2015-11-13
check if user is active in overview

learykara:
Goes along with https://github.com/Thinkful/cuckoo/pull/309 -- checks if a user is active in the `OpenSessionOverview` component before allowing them to RSVP.  Note this only applies to QA sessions (can't RSVP for a workshop from that overview screen)

### DIFF
--- a/src/OpenSessionOverview/OpenSessionOverview.jsx
+++ b/src/OpenSessionOverview/OpenSessionOverview.jsx
@@ -182,6 +182,8 @@ class OpenSessionOverview extends React.Component {
     } = this.props;
     const {detail_page_url, id} = session;
 
+    const userIsActive = !_.isEmpty(user) && user.account_status === 'active';
+
     return (
       <div className="qa-session-ctas">
         {linkToCalendar ?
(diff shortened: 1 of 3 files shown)


src/OpenSessionOverview/OpenSessionOverview.jsx
```diff
@@ -192,7 +192,7 @@ class OpenSessionOverview extends React.Component {
             <Icon name="navigateright" className="button-right-icon"/>
           </a>
         : (
-          !user ?
+          !user.isActive() ?
```
bgribble: it seems a bit weird to me that thinkful-ui is depending on a function (`isActive`) that's just defined in cuckoo.   Shouldn't that be in the common lib?

bgribble:
Other than the comment above, :+1:

src/OpenSessionOverview/OpenSessionOverview.jsx
learykara: yeah good point. since the user model isn't defined in ui yet I'll just check for the account_status here directly

PR #234, opened 2015-11-17
TopicPicker Relocates to UI

benjaminEwhite:
Note:

* This required small changes to `Tag` component
* Fixes a bug in `_filterTopicList` whereby inputting a topic like `C++` when `availableTopics` was passed in caused an error.
* Makes `minTopicLength` configurable, but defaults to 0.
* Has new requirements so running locally will need to ./script/install

Also, shortly after this is released, both Raven and Cuckoo will be updated to comply with updated `Tag` props, and to use `TopicPicker` from UI.

### DIFF
--- a/client/DemoPage.jsx
+++ b/client/DemoPage.jsx
@@ -12,7 +12,8 @@ const {
   Icon,
   Loader,
   NotificationView,
-  OpenSessionOverview
+  OpenSessionOverview,
+  TopicPicker
 } = require('../src')
 
 require('./styles/demo.less')
@@ -142,6 +143,15 @@ class DemoPage extends React.Component {
               config: Demo.props.json({projects: {url: ''}})
             }}/>
 
+            <h3>TopicPicker</h3>
+            <Demo
+              target={TopicPicker}
+              props={{
+                addMatchEmphasis: Demo.props.bool(true),
+                minTopicLength: Demo.props.constant(3),
+                activeTopics: Demo.props.json(['HTML', 'CSS']),
+                availableTopics: Demo.props.json(
+                  ['HTML', 'CSS', 'React', 'Node'])}}/>
           <RouteHandler />
         </div>
         <Footer/>
--- a/client/DemoPage.jsx
+++ b/client/DemoPage.jsx
@@ -13,7 +13,7 @@ const {
   Loader,
   NotificationView,
   OpenSessionOverview,
-  TopicPicker
+  TopicPicker,
 } = require('../src')
 
 require('./styles/demo.less')
@@ -147,10 +147,15 @@ class DemoPage extends React.Component {
             <Demo
               target={TopicPicker}
               props={{
+                activeTopics: Demo.props.json(['foo', 'bar']),
                 addMatchEmphasis: Demo.props.bool(true),
-                activeTopics: Demo.props.json(['HTML', 'CSS']),
                 availableTopics: Demo.props.json(
-                  ['HTML', 'CSS', 'React', 'Node'])}}/>
+                  ['HTML', 'hockey', 'horses', 'hypervisor']),
+                className: Demo.props.constant('topic-picker-demo-page'),
+                handleUpdateTopics: Demo.props.callback.log,
+                maxSuggestions: Demo.props.choices([2, 3, 4]),
+                minTopicLength: 3,
+              }}/>
           <RouteHandler />
         </div>
         <Footer/>
--- a/client/DemoPage.jsx
+++ b/client/DemoPage.jsx
@@ -151,7 +151,6 @@ class DemoPage extends React.Component {
                 addMatchEmphasis: Demo.props.bool(true),
                 availableTopics: Demo.props.json(
                   ['HTML', 'hockey', 'horses', 'hypervisor']),
-                className: Demo.props.constant('topic-picker-demo-page'),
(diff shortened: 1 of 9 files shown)


src/Tag/Tag.jsx
```diff
@@ -2,12 +2,21 @@ const cx = require('classnames');
...
+
+  static propTypes = {
+    className: React.PropTypes.string,
+    displayName: React.PropTypes.string,
+    url: React.PropTypes.string
+  }
+
   render() {
-    const {children, className, config, displayName} = this.props;
+    const {children, className, displayName, url} = this.props;
```
ripleyaffect: this change is going to break functionality in cuckoo and raven. make sure you connect w/ @kara for this. Maybe just default to using config/displayName to create url if it's not passed explicitly.

benjaminEwhite: Already discussed with @learykara. See notes at top of PR -- will follow up release of this with releases to Cuckoo and Raven (which I'll test locally with updated UI first).

src/TopicPicker/TopicPicker.jsx
```diff
@@ -0,0 +1,283 @@
...
+    minTopicLength: DEFAULT_MIN_TOPIC_LENGTH,
+    // if parent doesn't pass in callback, to avoid conditionals inline
+    handleUpdateTopics: () => null,
+  }
+
+  componentDidMount() {
+    const {activeTopics} = this.props;
+    this.setState({topics: activeTopics});
+
+    ReactDOM.findDOMNode(this.refs.topicForm).
```
ripleyaffect: This ref now refers to the dom node itself, so this isn't needed.

src/TopicPicker/TopicPicker.jsx
```diff
@@ -0,0 +1,283 @@
...
+      addEventListener('keydown', e => this._handleKeyDown(e));
```
ripleyaffect: This anonymous function isn't necessary, you can just pass `this._handleKeyDown`. This will actually cause an error if it is remounted, because the previous listener will still be around, since the `removeEventListener` in `componentWillUnmount` tries to remove a new anonymous function.

src/TopicPicker/TopicPicker.jsx
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  componentWillUnmount() {
+    ReactDOM.findDOMNode(this.refs.topicForm).
```
ripleyaffect: Ditto on not needing this `findDOMNode` call

src/TopicPicker/TopicPicker.jsx
```diff
@@ -0,0 +1,283 @@
+const cx = require('classnames');
+const escapeStringRegexp = require('escape-string-regexp');
+const marked = require('marked');
+const React = require('react');
+const ReactDOM = require('react-dom');
+
+const {Icon} = require('../Icon');
+const {Tag} = require('../Tag');
+
+
```
ripleyaffect: xtra spacer

src/TopicPicker/TopicPicker.jsx
```diff
@@ -0,0 +1,283 @@
...
+    let topic;
+
+    if (selectedSuggestionIndex > -1) {
+      topic = this._filterTopicList()[selectedSuggestionIndex];
+    }
+    else {
+      topic = this.refs.topicInput.value;
+    }
+
+    if (topic.length > this.props.minTopicLength) {
```
ripleyaffect: this should be >= and the default should be 1 I suppose. "minimum" indicates the minimum valid value

client/DemoPage.jsx
```diff
@@ -142,6 +143,14 @@ class DemoPage extends React.Component {
...
             }}/>
 
+            <h3>TopicPicker</h3>
+            <Demo
+              target={TopicPicker}
+              props={{
+                addMatchEmphasis: Demo.props.bool(true),
+                activeTopics: Demo.props.json(['HTML', 'CSS']),
+                availableTopics: Demo.props.json(
+                  ['HTML', 'CSS', 'React', 'Node'])}}/>
```
ripleyaffect: Why not add all the options here in the demo?

client/DemoPage.jsx
```diff
@@ -12,7 +12,8 @@ const {
   Icon,
   Loader,
   NotificationView,
-  OpenSessionOverview
+  OpenSessionOverview,
+  TopicPicker
```
ripleyaffect: I usually add the comma on the last item, since es6 allows it and it makes moving/adding/removing items much simpler

src/TopicPicker/index.es6
```diff
@@ -0,0 +1,5 @@
+const {TopicPicker} = require('./TopicPicker');
```
ripleyaffect: you can also just `export default require('./TopicPicker')`, which does the same thing (exports the object)

ripleyaffect:
Need to bump the version of ui, then :+1:

src/Tag/Tag.jsx
ripleyaffect: :ok_hand:

src/TopicPicker/TopicPicker.jsx
benjaminEwhite: yeah, silly mistake.

PR #236, opened 2015-11-19
Check for undefined "to" and "from" args

bgribble:
Fixes the error thrown for some secret signup enrollments when no "from" address is passed to AnalyticsAPI.alias. 

This addresses https://github.com/Thinkful/Thinkful/issues/154

### DIFF
--- a/src/analytics/actions.es6
+++ b/src/analytics/actions.es6
@@ -161,8 +161,8 @@ function alias(to, from, options, fn) {
     if (is.object(from)) options = from, from = null;
 
     // Aliasing Thinkful emails is dangerous, as we impersonate
-    if (to.indexOf('@thinkful.com') == -1 &&
-            from.indexOf('@thinkful.com') == -1) {
+    if (to && to.indexOf('@thinkful.com') == -1 &&
+            (! from || from.indexOf('@thinkful.com') == -1)) {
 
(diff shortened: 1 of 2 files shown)


src/analytics/actions.es6
```diff
@@ -161,8 +161,8 @@ function alias(to, from, options, fn) {
     if (is.object(from)) options = from, from = null;
 
     // Aliasing Thinkful emails is dangerous, as we impersonate
-    if (to.indexOf('@thinkful.com') == -1 &&
-            from.indexOf('@thinkful.com') == -1) {
+    if (to && to.indexOf('@thinkful.com') == -1 &&
+            (! from || from.indexOf('@thinkful.com') == -1)) {
```
learykara: if there is no `from` then why bother aliasing?

learykara:
I think we want the opposite behavior-- if no `from` is supplied then aliasing doesn't actually do anything

other than that, :+1:

src/analytics/actions.es6
bgribble: This is a newly created email, so I guess the idea is to associate this user with the new email.  I'm not sure I understand the differences between "identify" and "alias" in this case.   In any case, the code should not blow up with no `from`.

PR #238, opened 2015-11-22
get rid of use of refs and use onKeyDown

benjaminEwhite:
topic picker was failing in some apps with error
discussed here: https://gist.github.com/jimfb/4faa6cbfb1ef476bd105

after numerous dead ends, turned out that the refs inside
TopicPicker were causing this issue. Easily addressed by getting
rid of refs.

### DIFF
--- a/src/TopicPicker/TopicPicker.jsx
+++ b/src/TopicPicker/TopicPicker.jsx
@@ -56,6 +56,7 @@ class TopicPicker extends React.Component {
     activeTopics: [],
     maxSuggestions: 10,
     minTopicLength: DEFAULT_MIN_TOPIC_LENGTH,
+    placeholderText: "Add a tag (hit 'return' after each one)",
     // if parent doesn't pass in callback, to avoid conditionals inline
     handleUpdateTopics: () => null,
   }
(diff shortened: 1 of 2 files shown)


src/TopicPicker/TopicPicker.jsx
```diff
@@ -224,6 +221,7 @@ class TopicPicker extends React.Component {
 
   render() {
     const {pattern, topics, selectedSuggestionIndex, isFocused} = this.state;
+    const {className, placeholderText} = this.props;
```
learykara: `className` unused (replace `this.props.className` on line 229)

src/TopicPicker/TopicPicker.jsx
```diff
@@ -34,6 +34,7 @@ class TopicPicker extends React.Component {
     this.state = {
       pattern: '',
       topics: [],
+      topicFormInputValue: null,
```
learykara: unused

learykara:
lgtm :+1:

src/TopicPicker/TopicPicker.jsx
benjaminEwhite: woops!

src/TopicPicker/TopicPicker.jsx
benjaminEwhite: remnant of earlier attempt to solve this bug. thanks.

PR #243, opened 2015-11-25
Flash cyber weekend sale

tholex:


demo/__env.json
```diff
@@ -149,14 +149,5 @@
...
     "hotload": true
-  },
-  "user": {
-    "tf_login": "tholex@gmail.com",
-    "name": "Abe Z",
-    "first_name": "Abe",
-    "last_name": "X",
-    "id": 1,
-    "contact_id": 1,
-    "access": ["navbar-notifications"]
```
diezguerra: are you sure this doesn't break other things?

src/AppBar/Navigation.jsx
```diff
@@ -64,6 +65,12 @@ class AppNav extends React.Component {
         }, 400);
     }
 
+    renderFlash() {
+        return (<div href="//www.thinkful.com/enroll/" className="app-nav-flash">
```
learykara: `config.enroll.url`?

src/AppBar/Navigation.jsx
```diff
@@ -202,6 +209,14 @@ class AppNav extends React.Component {
         const navClassName = cx(
             'app-nav', {'app-nav__visible': this.state.isMenuVisible});
 
+        const blackFriday = moment("2015-11-27T00:00:00Z");
+        const cyberTuesday = moment("2015-12-01T12:00:00Z");
+        const current = moment();
```
learykara: `moment.tz('UTC')`

src/AppBar/Navigation.jsx
```diff
@@ -202,6 +209,14 @@ class AppNav extends React.Component {
...
+        /* flashActive is true between Nov 27, 00:00 GMT, and
```
learykara: shouldn't it be true during those times in the user's timezone?

demo/__env.json
tholex: reverted to keep demo page as-is

src/AppBar/Navigation.jsx
tholex: turns out it's just not going to be linked, forgot to remove the href.

src/AppBar/Navigation.jsx
tholex: Pretty sure this is OK? We'll start the sale when the UK hits black friday (7pm EST on Thanksgiving) and end it when japan hits midnight onDec 1st (~7am EST Dec 1st)? Asking justin

PR #244, opened 2015-11-25
Feature/one click copy

learykara:
move the component from settings.tf.com/refer-a-friend w one-click-copy to thinkful-ui

### DIFF
--- a/src/OneClickCopy/OneClickCopy.jsx
+++ b/src/OneClickCopy/OneClickCopy.jsx
@@ -11,6 +11,11 @@ class OneClickCopy extends React.Component {
     onCopyClickFail: React.PropTypes.func,
   }
 
+  _handleInputClick = (event) => {
+    event.preventDefault();
+    this.refs.copyArea.select();
+  }
+
   _handleCopyToClipboard = (event) => {
     const {onCopyClick, onCopyClickSuccess, onCopyClickFail} = this.props;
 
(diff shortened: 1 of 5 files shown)


src/OneClickCopy/OneClickCopy.jsx
```diff
@@ -0,0 +1,56 @@
...
+            [React.PropTypes.string, React.PropTypes.object]),
+    inputText: React.PropTypes.string,
+    onCopyClick: React.PropTypes.func,
+    onCopyClickSuccess: React.PropTypes.func,
+    onCopyClickFail: React.PropTypes.func,
+  }
+
+  _handleInputClick = (event) => {
+    event.preventDefault();
+    this.refs.copyArea.select();
```
ripleyaffect: why this selection behavior, if it's just duplicated in the `_handleCopyToClipboard` function?

src/OneClickCopy/index.es6
```diff
@@ -0,0 +1,5 @@
+const {OneClickCopy} = require('./OneClickCopy');
+
+require('./one-click-copy.less');
+
+module.exports = {OneClickCopy}
```
ripleyaffect: `export default require('./OneClickCopy');` does the same thing

ripleyaffect:
:+1:

src/OneClickCopy/OneClickCopy.jsx
learykara: if you click in the input at all, the whole thing should be selected but not editable. so this is separate from the copy function

src/OneClickCopy/index.es6
learykara: will update

PR #245, opened 2015-11-30
no refs, man

learykara:


benjaminEwhite:
Reluctantly thumbsupping ( :+1: ) cause one should never have to do `event.target.parentElement.firstChild.select()`, but for now, the show must go on.

PR #246, opened 2015-11-30
Alias react

benjaminEwhite:
This appears to solve the dreaded invariant violation error we've been getting when using components with `.refs` from thinkful-ui (discussed [here](https://gist.github.com/jimfb/4faa6cbfb1ef476bd105)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "thinkful-ui",
-  "version": "2.0.14",
+  "version": "2.0.15",
   "description": "Shared navigation and UI resources for Thinkful.",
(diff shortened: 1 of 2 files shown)


benjaminEwhite:
Still need to increment package.json version, but holding off because there are several outstanding PRs for UI atm, and will update just before merging.

bgribble:
this appears to help in at least some cases, i.e. Owl, and at least doesn't (I think) hurt anything, so :+1:

not exactly a ringing endorsement :) but it's the best I can do for first-hand knowledge

PR #253, opened 2015-12-02
Add app info to identify calls.

rkuykendall:


src/analytics/actions.es6
```diff
@@ -149,6 +156,8 @@ function identify(id, traits, options, fn) {
         }
     }
 
+    traits = mergeIntoDict(traits, appInfo);
```
ripleyaffect: This `mergeIntoDict` function is just [`_.defaults`](https://lodash.com/docs#defaults) or [`_.assign`](https://lodash.com/docs#assign). using one of those is preferred

ripleyaffect:
:+1:

PR #255, opened 2015-12-02
Add LinkWithSlug component

ripleyaffect:


ripleyaffect:
This doesn't add `LinkWithSlug` to the index of `src` because I do not want to include react-router@1.0.0. This requires imports of `LinkWithSlug` to be done as
```
const { LinkWithSlug } = require('thinkful-ui/src/LinkWithSlug')
```
until the other apps are upgraded to use react-router 1.0.0

bgribble:
ok lgtm :+1:

PR #261, opened 2015-12-18
Switch to _.get for more secure variable checking.

rkuykendall:


src/analytics/actions.es6
```diff
@@ -1,18 +1,21 @@
 const log = require('debug')('ui:analytics');
-const result = require('lodash/object/result');
-const defaults = require('lodash/object/defaults');
 const Qs = require('qs');
 const is = require('is');
 const superagent = require('superagent');
 
+const defaults = require('lodash/object/defaults');
+const get = require('lodash/object/get');
```
ripleyaffect: I think this can be consolidated into `const { defaults, get } = require('lodash/object')`

src/analytics/actions.es6
```diff
@@ -91,26 +94,23 @@ function load(writeKey) {
...
+        writeKey = meta && get(meta, 'dataset.token');
     }
 
     // Select from __env
-    if (!writeKey) {
-        if (__env && __env.config.vendor.segment.token) {
-            writeKey = __env.config.vendor.segment.token;
-        }
+    if (!writeKey && get(__env, 'config.vendor.segment.token')) {
+        writeKey = __env.config.vendor.segment.token;
```
ripleyaffect: this could be `writeKey = writeKey || get(__env, 'config.vendor.segment.token')`

src/analytics/actions.es6
rkuykendall: I'll try it. Will also do the other two in that form, just since it reads clean and I wouldn't be surprised if I added more in the future.

ripleyaffect:
:+1:

src/analytics/actions.es6
rkuykendall: ooooo, interesting. I like it a lot, but considering the 3 around it are in the form `if (!writeKey)` and can't be shortened, I think leaving them all the same makes it easier to read the code and understand what's happening.

src/analytics/actions.es6
rkuykendall: Hm, sadly it doesn't look like it can be.

PR #265, opened 2016-01-04
use avatar endpoint for app bar image

learykara:
needs to follow https://github.com/Thinkful/pysplash/pull/2345

### DIFF
--- a/src/AppBar/Navigation.jsx
+++ b/src/AppBar/Navigation.jsx
@@ -128,6 +128,7 @@ class AppNav extends React.Component {
                         <span alt="Menu" className="app-nav-burger"></span>
                         <Gravatar
                             className="app-nav-gravatar"
+                            email=""
                             src={`${config.api.url}/api/hupers/me/avatar`}
                             size={120}/>
                     </a>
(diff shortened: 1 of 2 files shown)


src/AppBar/Navigation.jsx
```diff
@@ -126,7 +126,11 @@ class AppNav extends React.Component {
                     <Notifications />
                     <a className="app-nav-link app-nav-link__toggle" onClick={this._toggleMenu}>
                         <span alt="Menu" className="app-nav-burger"></span>
-                        <Gravatar className="app-nav-gravatar" email={user.tf_login} size={120}/>
+                        <Gravatar
+                            className="app-nav-gravatar"
+                            email=""
+                            src={`${config.api.url}/api/hupers/me/avatar`}
```
tholex: why use the Gravatar tag if it's not fetching a gravatar? `img` should do.

tholex: the class is what determines placement/size etc

learykara: there are already a bunch of places throughout the app where we use the `Gravatar` component even though we use an explicit `src` and not email -- figured if nothing else it makes it easier to quickly find all of the places where we show peoples avatars as opposed to other images. seem a good enough reason to keep it?

tholex: weird that `api.url` comes without `/api` but it's correct :+1:

tholex:
I guess that's alright – np :+1:

PR #268, opened 2016-01-20
Protect identify and alias from being fired where they shouldn't be.

rkuykendall:


.eslintrc
```diff
@@ -195,7 +195,7 @@
     "quotes": [1, "single"],        // specify whether double or single quotes should be used
     "semi": 0,                      // require or disallow use of semicolons instead of ASI
     "sort-vars": 0,                 // sort variables within the same declaration block (off by default)
-    "space-after-function-name": 1, // require a space after function names (off by default)
+    "space-before-function-paren": 0, // require a space after function names (off by default)
```
diezguerra: hm

diezguerra:
:+1:

.eslintrc
rkuykendall: It's a change in the linter, this is the new name for this rile. And also, we never use this rule so I don't know why we had it set to required. Also I looked through a bunch of our newer code and everyone writes `myFunction() { }` which makes perfect sense to me, so I just turned this off.

rkuykendall:
Marking this ready to :ship: after testing solution with Catherine today.

.eslintrc
diezguerra: :+1:

src/analytics/actions.es6
```diff
@@ -20,6 +20,20 @@ const cookies = zipObject(map(document.cookie.split('; '), function(cookie) {
...
+
+function isImpersonating() {
+  let retVal = _.has(window.__env, 'user.real_admin_tf_login');
+
+  if (retVal) {
+    log('No analyitcs for impersonating users.');
+  }
+
+  return retVal;
+}
```
diezguerra: this might lead to believe the return type isn't boolean. maybe change the constant name to something else?

src/analytics/actions.es6
```diff
@@ -128,23 +142,43 @@ function load(writeKey) {
...
-    traits = defaults(traits || {}, appInfo);
+    // If logged in, always identify by user email
+    if (isLoggedIn()) {
+      id = __env.user.tf_login;
+
+    // If logged out, set the id to the mixpanel ID,
+    // unless Hawk provides one on login
+    } else {
+      if (appInfo.app == 'hawk' && id) {
+        // Keep Hawk-supplied ID
```
diezguerra: might make sense to explicitly say in the comment that this is a no op (e.g.  `// Do nothing`)

src/analytics/actions.es6
```diff
@@ -189,6 +235,10 @@ function track(event, properties, options, fn) {
 
 // This event mirrors the call signature of global.analytics.page
 function page(category, name, properties, options, fn) {
+    if (isImpersonating()) {
+      return;
+    }
```
diezguerra: I wonder if there isn't a higher level function where we can put this instead of repeating it -- nbd though

diezguerra:
:+1:

src/analytics/actions.es6

src/analytics/actions.es6

rkuykendall:
Marking ready to :ship:. Deploying right before the other PRs.

PR #269, opened 2016-01-22
add draft notice to top of open session overview

learykara:
image


src/OpenSessionOverview/OpenSessionOverview.jsx
```diff
@@ -123,6 +123,9 @@ class OpenSessionOverview extends React.Component {
         </div>
 
         <div className="session-overview-wrapper">
+          {!session.published_asof &&
+            <div className="subheading subheading__draft">Draft</div>
```
ripleyaffect: This isn't a type of subheading, it's a instance of one, with unique properties, so I think something like "draft-notice" is a better class name. the dunder pattern is for type modifications, overriding or adding properties to the subheading class, but this is a unique case.

ripleyaffect:
:+1:

src/OpenSessionOverview/OpenSessionOverview.jsx
learykara: gotcha makes sense

PR #272, opened 2016-01-26
add /responsible-disclosure to footer

tati:
Add a Responsible Disclosure Program link to in-app footer https://www.thinkful.com/responsible-disclosure/


src/Footer/Footer.jsx
```diff
@@ -47,6 +47,10 @@ function generateLinkSet(config) {
     {
       'name': 'Blog',
       'location': `//blog.thinkful.com`
+    },
+    {
+      'name': 'Responsible Disclosure Program',
+      'location': `${config.www.url}/responsible-disclosure`
```
diezguerra: no / ?

diezguerra:
:+1: minus comments

src/Footer/Footer.jsx
tati: none of the other ones have a trailing slash :question: @learykara?

learykara: they should - we use trailing slashes everywhere (except, I've noticed, ostrich)

tati:
fixed in dif PR

PR #273, opened 2016-01-27
Safer mixpanel check, send identity to Oilbird, Pelican can alias.

rkuykendall:


src/analytics/actions.es6
```diff
@@ -58,6 +58,27 @@ function tryEmail() {
...
+    return id;
+  }
+
+  // If logged out, set the id to the mixpanel ID,
+  if (is.function(_.get(window, 'mixpanel.get_distinct_id'))) {
+    return window.mixpanel.get_distinct_id();
+  }
+
+  // If we can't find mixpanel, fallback to null ID
+  return id;
```
diezguerra: how is this null? this could be anything, potentially -- whatever is passed

src/analytics/actions.es6
```diff
@@ -194,8 +200,8 @@ function alias(to, from, options, fn) {
     }
 
     // See Mixpanel rules in identify function
-    if (appInfo.app != 'tailorbird') {
-        log('Alias should only be called on account creation.');
+    if (appInfo.app != 'tailorbird' && appInfo.app != 'pelican') {
```
diezguerra: make sure pelican has appInfo.app and it equals to that, if you haven't. This convention isn't spread throughout the repos I suspect

diezguerra: make sure pelican has appInfo.app and it equals to that, if you haven't. This convention isn't spread throughout the repos I suspect

diezguerra:
:+1: minus comments

src/analytics/actions.es6
rkuykendall: Yeah, I'm planning on checking that when I add these aliases. Right now they don't exist. Actually a lot of our apps are missing that info. It's very easy to add though.

src/analytics/actions.es6

PR #274, opened 2016-01-27
Upgrade to SG4.0, sidebar-friendly timezone display

tholex:


src/Footer/footer.less
```diff
@@ -17,20 +17,17 @@
 }
 
 .footer-timezone {
-  padding-left: 15px;
-  padding-right: 15px;
-  margin-bottom: 15px;
+  .proxima-light;
```
learykara: wait I thought we got rid of proxima light?

src/Footer/footer.less
```diff
@@ -17,20 +17,17 @@
...
+  background-color: @gray15;
```
learykara: why put a background color on this part? shouldn't it be transparent/ the same as whatever page it's included on?

src/Footer/footer.less
```diff
@@ -17,20 +17,17 @@
...
+  padding: 15px;
+  margin-bottom: 0;
+  text-align: center;
+  max-width: none;
 
-  @media @nonmobile {
-    padding-left: 0;
-    padding-right: 0;
-    margin-bottom: 30px;
+  @media @mobile {
```
learykara: `@media @nonmobile` is already in this file - should continue to use that, no?

learykara:
:+1: other than comments

learykara:
what will happen if new UI version is installed in repos that have sg 3.0?

src/Footer/footer.less
tholex: It's still there, it's just not the default

src/Footer/footer.less
tholex: I'd avoid mixing `mobile` and `nonmobile` on the same element, but it's fine in the same file. Basically if there's 2-3+ media queries on the same element, only the `nonmobile` and "going-up" ones should be used.

src/Footer/footer.less
tholex: In effect, it has a white background now, since it's part of the footer.

With the sidebar in dashboard and some negative margin hacks, it looks like this:


The gray bar should work better with either an all-white page or a page with a sidebar.

tholex: https://files.slack.com/files-pri/T026YFPQS-F0KFB00Q3/pasted_image_at_2016_01_26_07_41_pm.png

learykara: changing the `text-align` to `center` on L25 takes care of the sidebar / negative margin hacks issue doesn't it?

src/Footer/footer.less
learykara: ahhh I always thought that comment meant in the same file, not just on the same element. noted.

tholex:
Just checked the nav in an app with SG3 (and the old font file) – fonts look a little thin, but nothing to worry about. That's really the only concern I can think of.

src/Footer/footer.less
tholex: It would on most screen sizes, yes, but I think the footer with no top padding and white background under the TZ display is too restrictive.

PR #275, opened 2016-01-27
Small fix for edge case in identifying with traits as first parameter and no ID.

rkuykendall:
I don't believe we have any code in the wild that actually hits this, but I noticed it and it should be fixed anyways.

### DIFF
--- a/src/analytics/actions.es6
+++ b/src/analytics/actions.es6
@@ -170,7 +170,7 @@ function identify(id, traits, options, fn) {
     // Argument reshuffling, from original library.
     if (is.function(options)) fn = options, options = null;
     if (is.function(traits)) fn = traits, options = null, traits = null;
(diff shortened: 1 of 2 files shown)


src/analytics/actions.es6
```diff
@@ -170,7 +170,7 @@ function identify(id, traits, options, fn) {
     // Argument reshuffling, from original library.
     if (is.function(options)) fn = options, options = null;
     if (is.function(traits)) fn = traits, options = null, traits = null;
-    if (is.object(id)) options = traits, traits = id, id = user.id();
+    if (is.object(id)) options = traits, traits = id, id = window.analytics._user.id();
```
tholex: `global` instead of `window` ?

tholex:
:+1:

PR #276, opened 2016-01-28
Changes to allow for getting most up-to-date __env.

rkuykendall:
- Instead of `__env` and `appInfo` being consts, source them from the most up-to-date versions.
- Only use directly imported lodash functions.
- Tested with Pelican and it worked, but review closely since the changes are very spread out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "thinkful-ui",
-  "version": "2.0.41",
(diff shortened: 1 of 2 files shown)


bgribble:
:+1: looks good

bgribble:
Still looks good :+1: :+1:

PR #280, opened 2016-02-05
Tf-Guest signins see unauthed version of navbar

bgribble:
Tf-Guest signins are "Authenticated" so they have a `window.__env.user`, meaning AppBar will try to show the logged-in version.  Luckily `role !== 'guest'` will discriminate.


ripleyaffect:
:+1:

PR #282, opened 2016-02-09
Add className support for AppNav

ripleyaffect:


src/AppBar/Navigation.jsx
```diff
@@ -72,13 +73,14 @@ class AppNav extends React.Component {
...
     renderAuthed(user, config) {
+        const { className } = this.props;
         const navClassName = cx(
             'app-nav', {'app-nav__visible': this.state.isMenuVisible});
         const navLinks = linkSet.main.filter(link => !link.search);
         const searchLink = linkSet.main.filter(link => link.search)[0];
 
         return (
-            <div className='app-nav-container'>
+            <div className={cx('app-nav-container', className)}>
```
tati: so it will always have `app-nav` and then if menu is visible is will have `app-nav__visible`?

tati:
not sure if you want someone else to review but I think I understand it :grin: :+1:

src/AppBar/Navigation.jsx
ripleyaffect: if `className` is defined, it'll be on the `class` of the rendered `div`. if it's not defined, nothing will change. `navClassName` is a different variable used below on the `nav` tag

PR #284, opened 2016-02-23
Feature/stork analytics

learykara:


src/analytics/actions.es6
```diff
@@ -1,3 +1,5 @@
+'use strict';
+
```
ripleyaffect: Should be added by babel?

ripleyaffect:
👍

src/analytics/actions.es6
learykara: was getting an error in stork... something up with the fact that there's no jsx in there?

learykara: ah - that error was thrown before I added babel to stork's webpack. makes sense - removing

PR #288, opened 2016-03-17
Feature/remove paths and search

tati:
This PR should remove paths and search from all nav states.

Looks good locally for the admin view...any advice on how to check student view? Is there an email I can log into hawk with?

screen shot 2016-03-17 at 4 24 57 pm

### DIFF
--- a/src/AppBar/linkSet.es6
+++ b/src/AppBar/linkSet.es6
@@ -51,18 +51,20 @@ else {
         if (user.access.indexOf('core-student') >= 0) {
             defaults(home, config.dashboard);
             main.push(home);
+            main.push(config.workshops);
         }
         else if (/tfl/.test(user.student_type)) {
             assign(home, {
-                host: config.projects.host
+                host: config.projects.host,
+                url: config.workshops.url
             });
             main.push(home);
         }
         else {
             defaults(home, config.dashboard);
             main.push(home);
+            main.push(config.workshops);
         }
-        main.push(config.workshops);
     }
 
     menu.push(config.refer);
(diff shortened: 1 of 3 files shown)


src/AppBar/linkSet.es6
```diff
@@ -63,12 +51,10 @@ else {
         if (user.access.indexOf('core-student') >= 0) {
             defaults(home, config.dashboard);
             main.push(home);
-            main.push(config.paths);
         }
         else if (/tfl/.test(user.student_type)) {
             assign(home, {
                 host: config.projects.host,
-                url: config.paths.url
```
learykara: where are tfl students' home buttons supposed to lead?

learykara:
for looking at student view... why not just impersonate?

src/AppBar/linkSet.es6
tati: ah good catch, chatted with Varun and he says should go to workshops so I am removing workshops from the TFL nav but making the home button go there.

tati:
Where do you normally get the emails from to impersonate them? I am getting this error. Or is it something to do with my db...I did a sanitize backup recently 

screen shot 2016-03-17 at 5 14 08 pm

learykara:
you need `*()*` at the end of non-admin emails locally.

tati:
gotcha gracias...should be good to go now and I looked locally

learykara:
might be smart to also get rid of the hacky logic in the actual nav component that adds the search link (example [here](https://github.com/Thinkful/ui/blob/master/src/AppBar/Navigation.jsx#L86))

learykara:
:+1:

PR #290, opened 2016-03-22
allow dropdown to accept simple array

learykara:
The `Dropdown` component previously accepted an array of objects for data, along with a `selectedInd` for the selected item ala:
```
data = [
  {'value': 1, 'displayName': 'foo'},
  {'value': 2, 'displayName': 'bar'}];
```

This PR allows it to accept just a simple array of strings and assumes that those strings should be both the value and display name for each item.  It also now accepts a `value` prop, which will replace `selectedInd` so that you don't need to search for the index outside of the component

### DIFF
--- a/src/Dropdown/Dropdown.jsx
+++ b/src/Dropdown/Dropdown.jsx
@@ -16,7 +16,10 @@ require('./dropdown.less');
 const Dropdown = React.createClass({
 
   propTypes: {
-    data: React.PropTypes.array.isRequired,
+    data: React.PropTypes.arrayOf(
+      React.PropTypes.oneOfType([
+        React.PropTypes.string, React.PropTypes.object
+      ])).isRequired,
     initialSelectedInd: React.PropTypes.number,
     selectedInd: React.PropTypes.number,
     defaultDisplay: React.PropTypes.string,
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.jsx
```diff
@@ -53,8 +53,8 @@ const Dropdown = React.createClass({
           className={cx("dropdown-item", item.className)}
           id={ind}
           key={ind}
-          value={item.value}>
-          {item.displayName}
+          value={item.value || item}>
+          {item.displayName || item}
```
ripleyaffect: safer way to do this is to create an object out of item if it's a string, where both `value` and `displayName` are == `item`. Also add to propTypes

src/Dropdown/Dropdown.jsx
```diff
@@ -70,20 +70,52 @@ const Dropdown = React.createClass({
...
+    if (typeof(data[0]) === "object") {
+      // Case 3: data is array of objects with {displayName: 'foo', value: 'bar'}
+      const values = data.map(item => item.value);
+    }
+    else {
+      // Case 4: data is array of values
+      const values = data;
+    }
+
+    return data.indexOf(value);
```
ripleyaffect: `return _.map(data, item => item.value || item).indexOf(value)` can replace 82 - 95, though comments are good

src/Dropdown/Dropdown.jsx
```diff
@@ -70,20 +70,52 @@ const Dropdown = React.createClass({
...
+
+    if (selectedInd === -1) {
+      return defaultDisplay;
+    }
+
+    if (typeof(data[0]) === "object") {
+      return data[selectedInd] && data[selectedInd].displayName || defaultDisplay;
+    }
+
+    return data[selectedInd]
```
ripleyaffect: `return _.map(data, item => item.displayName || item)[selectedInd] || defaultDisplay`

ripleyaffect:
propTypes, otw :+1:

src/Dropdown/Dropdown.jsx
learykara: y u so much cleverer than me

PR #296, opened 2016-03-29
disable workshops link in onboarding

learykara:
Result:

image

### DIFF
--- a/src/AppBar/app-bar.less
+++ b/src/AppBar/app-bar.less
@@ -187,7 +187,7 @@
   }
 
   &.disabled {
-    color: mix(@gray50, @gray25, 40%);
+    color: @blue75;
     pointer-events: none;
   }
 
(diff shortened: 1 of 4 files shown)


src/AppBar/app-bar.less
```diff
@@ -186,6 +186,11 @@
     .proxima-bold;
   }
 
+  &.disabled {
+    color: mix(@gray50, @gray25, 40%);
```
tholex: I think this is `@gray40`

tholex:
:+1:

src/AppBar/app-bar.less
tholex: blue50 or blue75 might work well here

learykara: oh wow ya that looks way better. I was struggling to make it look not like poop

PR #297, opened 2016-03-29
show Q&A sessions in nav not workshops

tati:
As part of TFL clean up we want "Q&A sessions" in nav not "Workshops"

I also just added some comments about which view is which...let me know if you think I should remove

### DIFF
--- a/src/AppBar/linkSet.es6
+++ b/src/AppBar/linkSet.es6
@@ -11,6 +11,9 @@ if (global.__env) {
 let config = {
     workshops: {
         icon: 'users'
+    },
+    qaSessions: {
+        icon: 'users'
     }
 }
 if (global.__env) {
@@ -26,6 +29,7 @@ let menu = [];
 let insertCourseDropdown = false;
 
 if(! user) {
+  // non-logged in user
     defaults(home, config.www);
     insertCourseDropdown = true;
     menu.push(config.mentors);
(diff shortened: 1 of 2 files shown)


src/AppBar/linkSet.es6
```diff
@@ -35,12 +39,13 @@ if(! user) {
 }
 
 else {
+  // admin, mentor
```
ripleyaffect: odd indention here and above

src/AppBar/linkSet.es6
```diff
@@ -61,9 +68,10 @@ else {
             main.push(home);
         }
         else {
+          // CP student??
```
ripleyaffect: could be CP or full time CP

src/AppBar/linkSet.es6
```diff
@@ -11,6 +11,9 @@ if (global.__env) {
 let config = {
     workshops: {
         icon: 'users'
```
ripleyaffect: can delete

src/AppBar/linkSet.es6
tati: updated

src/AppBar/linkSet.es6
tati: updated

src/AppBar/linkSet.es6
tati: fixed

ripleyaffect:
:+1:

tati:
fiinalllly confirmed locally 

screen shot 2016-03-30 at 5 33 18 pm

PR #300, opened 2016-03-30
Material-style loader instead of 'loading...'

tholex:
similar to http://codepen.io/tholex/pen/GZvYGm but a little smaller

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "react-avatar-editor": "^3.2.0",
     "react-dom": "^0.14.2",
     "react-hot-loader": "^1.2.7",
-    "reflux": "^0.2.12",
+    "reflux": "^0.3.0",
     "remarkable-loader": "^0.1.0",
     "spark-md5": "0.0.5",
     "style-loader": "^0.12.3",
--- a/src/Loader/Loader.jsx
+++ b/src/Loader/Loader.jsx
@@ -7,16 +7,22 @@ class Loader extends React.Component {
   static displayName = "Loader"
 
   static propTypes = {
-    className: React.PropTypes.string
+    className: React.PropTypes.string,
+    height: React.PropTypes.string
   }
 
   render() {
+    let styleObj = {height: '60px'};
+    if (this.props.height) {
+      styleObj = {height: this.props.height };
+    }
+
     return <div
-      className={cx('tui-loader', this.props.className || "")}>
-      Loading
-      <span className="tui-loader-dot"></span>
-      <span className="tui-loader-dot"></span>
-      <span className="tui-loader-dot"></span>
+      className={cx('tui-loader', this.props.className || "")}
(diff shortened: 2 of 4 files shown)


src/Loader/Loader.jsx
```diff
@@ -7,16 +7,22 @@ class Loader extends React.Component {
...
   static propTypes = {
-    className: React.PropTypes.string
+    className: React.PropTypes.string,
+    height: React.PropTypes.string
   }
 
   render() {
+    let styleObj = {height: '60px'};
+    if (this.props.height) {
+      styleObj = {height: this.props.height };
```
learykara: why not `const styleObj = {height: this.props.height || '60px'};`?

learykara: or take care of it with `defaultProps`?

src/Loader/Loader.jsx
```diff
@@ -7,16 +7,22 @@ class Loader extends React.Component {
...
+    }
+
     return <div
-      className={cx('tui-loader', this.props.className || "")}>
-      Loading
-      <span className="tui-loader-dot"></span>
-      <span className="tui-loader-dot"></span>
-      <span className="tui-loader-dot"></span>
+      className={cx('tui-loader', this.props.className || "")}
```
learykara: don't think you need the fallback `""` here

src/Loader/Loader.jsx
```diff
@@ -7,16 +7,22 @@ class Loader extends React.Component {
...
+      style={styleObj}>
+      <svg className="tui-loader-inner" height="25" width="25">
+  <circle className="tui-loader-stroke" cx="25" cy="25" r="15" fill="none" strokeWidth="5" strokeMiterlimit="10" />
+</svg>
```
learykara: weird indentation from L23 - 25?

package.json
```diff
@@ -82,7 +82,7 @@
     "react-avatar-editor": "^3.2.0",
     "react-dom": "^0.14.2",
     "react-hot-loader": "^1.2.7",
-    "reflux": "^0.2.12",
+    "reflux": "^0.3.0",
```
learykara: should this be part of this PR? doesn't seem related to reflux at all

learykara:
:+1:

src/Loader/Loader.jsx
tholex: went w/ defaultProps

src/Loader/Loader.jsx
tholex: aye

src/Loader/Loader.jsx
tholex: feexed

package.json
tholex: wasn't able to `npm install` without this change

PR #302, opened 2016-03-31
allow null values in dropdown

learykara:
allow falsy items to be values in dropdown


ripleyaffect:
:+1:

ripleyaffect:
assuming propTypes are ok with this, since they have the final say :smile:

PR #305, opened 2016-04-11
Request gzipped svgs and optimize local svgs.

rkuykendall:


.gitignore
```diff
@@ -1,3 +1,5 @@
+.DS_Store
```
tati: lol was this actually a problem? too funny

tati:
I would def check this once locally before you merge but :+1: I can check tomorrow morning for you :grin:

.gitignore
rkuykendall: ¯\_(ツ)_/¯ I had some pop up in my `git status`. I tend to just add it to the `.gitignore`, since I don't trust myself not to `git add -A`

PR #309, opened 2016-05-13
Feature/new footer

learykara:
Adds new in-app footer

image

image

### DIFF
--- a/client/DemoPage.jsx
+++ b/client/DemoPage.jsx
@@ -158,7 +158,7 @@ class DemoPage extends React.Component {
               }}/>
           <RouteHandler />
         </div>
-        <Footer includeTz={true} />
+        <Footer user={{timezone: 'America/New_York'}} />
       </div>);
   }
 }
--- a/src/Footer/Footer.jsx
+++ b/src/Footer/Footer.jsx
@@ -1,52 +1,120 @@
+const cx = require('classnames');
 const React = require('react');
-const moment = require('moment');
-const jstz = require('jstz');
+
 const {Icon} = require('../Icon');
 
 require('./footer.less');
 
+
 function generateLinkSet(config) {
   return [
     {
-      'name': 'Mentors',
-      'location': `${config.www.url}/mentors/`
-    },
-    {
-      'name': 'Student reviews',
-      'location': `${config.www.url}/reviews/`
-    },
-    {
-      'name': 'Pricing',
-      'location': `${config.www.url}/pricing/`
-    },
-    {
-      'name': 'About us',
-      'location': `${config.www.url}/about/`,
-      'subLink':
+      'heading': 'Courses',
+      'links': [
         {
-          'name': "We're hiring!",
-          'location': `${config.www.url}/about/#opportunities`
+          'name': 'Part Time Career Path',
+          'location': `${config.www.url}/courses/web-development-career-path/`,
+          'mobile': false
+        },
+        {
+          'name': 'Full Time Career Path',
+          'location': `${config.www.url}/courses/full-time-career-path/`,
+          'mobile': false
+        },
+        {
+          'name': 'Explore all courses',
+          'location': `${config.www.url}/courses/`,
+          'mobile': true
+        },
+        {
+          'name': 'Corporate training',
+          'location': `${config.www.url}/training-for-teams/`,
+          'mobile': true
+        },
+        {
+          'name': 'Bootcamp prep',
+          'location': `${config.www.url}/bootcamp-prep/`,
+          'mobile': false
+        },
+        {
+          'name': 'Pricing',
+          'location': `${config.www.url}/pricing/`,
+          'mobile': true
         }
+      ]
     },
     {
-      'name': 'Learning resources',
-      'location': `${config.www.url}/learn/`
+      'heading': 'Education',
+      'links': [
+        {
+          'name': '1-on-1 mentorship',
+          'location': `${config.www.url}/mentorship/`,
+          'mobile': false
+        },
+        {
+          'name': 'Career prep',
+          'location': `${config.www.url}/career-prep/`,
+          'mobile': false
+        },
+        {
+          'name': 'Job guarantee',
+          'location': `${config.www.url}/career-path-job-guarantee/`,
+          'mobile': false
+        },
+        {
+          'name': 'Student outcomes',
+          'location': `${config.www.url}/bootcamp-job-stats/`,
+          'mobile': false
+        },
+        {
+          'name': 'Bootcamp Finder',
+          'location': `${config.www.url}/bootcamps/`,
+          'mobile': true
+        },
+        {
+          'name': 'Learning resources',
+          'location': `${config.www.url}/learn/`,
+          'mobile': false
+        }
+      ]
     },
     {
-      'name': 'Training for teams',
-      'location': `${config.www.url}/training-for-teams/`
-    },
-    {
-      'name': 'Bootcamp prep',
-      'location': `${config.www.url}/bootcamp-prep/`
-    },
-    {
-      'name': 'Workshops and Q&A sessions',
-      'location': `${config.officeHours.url}/workshops/`
-    },
-    {
-      'name': 'Blog',
-      'location': `//blog.thinkful.com`
+      'heading': 'About',
+      'links': [
+        {
+          'name': 'Mission',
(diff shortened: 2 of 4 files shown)


learykara:
@darrellsilver @jcudmore what do you think?

darrellsilver:
I'm 👍; can't wait!

client/DemoPage.jsx
```diff
@@ -158,7 +158,7 @@ class DemoPage extends React.Component {
               }}/>
           <RouteHandler />
         </div>
-        <Footer includeTz={true} />
+        <Footer user={{timezone: 'America/New_York'}} />
```
ripleyaffect: could pull user out of context, as in most other components

jcudmore:
👍 😄

darrellsilver:
- Unfort I think we need to link to the homepage... students who hear about a redesign, or otherwise want to poke around, need _some_ way to do so. How about adding "Home" to the "About" column.
 - The 2nd & 3rd columns should be in alphabetical order, no?

jcudmore:
isn't the way to see the homepage just to log out? that's what I would do.

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
+
 const {Icon} = require('../Icon');
 
 require('./footer.less');
 
+const FACEBOOK_URL = 'https://www.facebook.com/thinkfulschool';
+const TWITTER_URL = 'https://twitter.com/thinkful';
+
+
 function generateLinkSet(config) {
```
ripleyaffect: `linkSet` is a bit of a misnomer. this is now returning sections with headings and smaller "link sets"

darrellsilver:
ya good point; updated!

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
+          'mobile': false
+        },
+        {
+          'name': 'Mentors',
+          'location': `${config.www.url}/mentors/`,
+          'mobile': false
+        },
+        {
+          'name': 'Site security',
+          'location': `${config.www.url}/responsible-disclosure/`,
```
ripleyaffect: this is the only thing you're using `config` for. You could pull it out at the top of the function to clean it up a little

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
-    includeTz: React.PropTypes.bool
-  }
-
-  static defaultProps = {
-    includeTz: false
   }
 
   render() {
-    const {includeTz, user} = this.props;
+    const { user={} } = this.props;
```
ripleyaffect: use `defaultProps` for this

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
+          'mobile': false
+        }
+      ]
     }
   ];
 }
 
 class Footer extends React.Component {
   static propTypes = {
     config: React.PropTypes.object,
```
ripleyaffect: `user` missing

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
-  }
-
-  static defaultProps = {
-    includeTz: false
   }
 
   render() {
-    const {includeTz, user} = this.props;
+    const { user={} } = this.props;
     const config = this.props.config || global.__env.config;
```
ripleyaffect: default props for this too and destructure with `user`

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
-          <div className="footer-centered">
-            <div className="footer-site-links">
-              <a
-                  className="button button__white"
-                  href={`${config.accounts.url}/login`}>
-                Sign in
+        <footer className="footer">
+          <div className="timezone timezone__mobile">
+            All times are in {user.timezone}
+          </div>
```
ripleyaffect: `{user.timezone && ...}` so the footer works for unauthed?

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
-                  <Icon name="twitter"/>
-                </a>
-              </div>
             </div>
-            <div className="footer-contact-links">
-              <a
-                  className="footer-main-link mobile-hidden"
-                  href="mailto:hello@thinkful.com">
-                Email us: hello@thinkful.com
+            {linkSet.map(section => <div className="footer-column">
```
ripleyaffect: `footer-column` should be it's own component with prop types etc. Will make it better when changing link sets around if that component grumbles about a missing prop on `section`. Also :+1: for calling this `section r.e. my comment on the name `linkSet` above

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
+                <h4 className="footer-heading">{section.heading}</h4>
+                {section.links.map(link => (
```
ripleyaffect: SectionLink should also be it's own component. you could use a functional component, though I tend to stick with class components because it standardizes the look of things like propTypes and grabbing things out of the context

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
+                  <a
+                      className={cx(
+                        "footer-link",
+                        {icon: link.icon, mobileHidden: !link.mobile})}
+                      href={link.location}>
+                    {link.icon ?
```
ripleyaffect: I tend to use `iconName` instead of `icon` being a string, seems weird to me that `icon` doesn't really mean an icon component /shrug up 2 u tho

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
+              </div>)
+            }
+            <div className="support-mobile">
+              <a className="footer-link" href={`${config.www.url}/support`}>
+                Support
               </a>
+            </div>
+          </div>
+          <div className="timezone">
+            All times are in {user.timezone}&nbsp;&nbsp;
```
ripleyaffect: `nbsb`s for spacing is putting a little bit of the view layer in here, like `<br>` tags. hashtag sketchy

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
+            <a href={`${config.settings.url}/profile`}>Change</a>
+          </div>
+          <div className="legal-links">
+            <span className="margin-span copyright">&copy; {moment().format('YYYY')} Thinkful, Inc.</span>
+            <span className="margin-span middot-desktop">&nbsp;·&nbsp;</span>
```
ripleyaffect: ditto r.e. nbsps, esp when there's a class on the span that you can give margins too

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
+              <a className="footer-link" href={`${config.www.url}/support`}>
+                Support
               </a>
+            </div>
+          </div>
+          <div className="timezone">
+            All times are in {user.timezone}&nbsp;&nbsp;
+            <a href={`${config.settings.url}/profile`}>Change</a>
+          </div>
+          <div className="legal-links">
```
ripleyaffect: also probably a good candidate for a component, esp as everything here is basically html. keeping components small just makes them easy to reason about and update. mentally, you get to do something similar to a BST search when looking for the thing you need to update.

src/Footer/Footer.jsx
```diff
@@ -1,135 +1,208 @@
...
   }
 
   render() {
-    const {includeTz, user} = this.props;
+    const { user={} } = this.props;
     const config = this.props.config || global.__env.config;
     const linkSet = generateLinkSet(config);
 
     return (
       <div className="footer-container">
```
ripleyaffect: don't think footer-container is necessary anymore? was useful when the timezone thing was there, but no longer

ripleyaffect:
saxy 🎷

:+1: after comments

src/Footer/Footer.jsx
learykara: yep - had thought that it wasn't possible to see this footer unauthed but I forgot about raven

src/Footer/Footer.jsx
learykara: needed to set the background color on the footer

client/DemoPage.jsx
learykara: are there not any lingering apps that don't have user in context?

Issue #314, opened 2016-06-10
Consider: Make Dropdown Accessible


jpdevries:
The current div based implementation of the `<Dropdown>` is not accessible to keyboards and screen readers. It also does not support native features like "autocomplete" that allow you to type the first few characters of an option to select it.

PR #315, opened 2016-06-10
Making Dropdown More Accessible

jpdevries:
Makes `<Dropdown>` more accessibleBy using native `<select>`, `<optgroup>`, and `<option>`.
- Now more keyboard accessible (supports "typeahead") and arrow keys to navigate.
- Screen Reader friendly.
- Supports nested optgroups (kinda).  

Closes Thinkful/ui#314

_Note: Build files are not checked in. Please run `webpack` after `git pull`._
### Pros
- Accessibility
- Keyboard control
- Native UX (mobile)
- more red than green (-90 + 71)
### Cons
- Can't style `<option>`

### Example Data

Data can be `optgroup` wrapper objects which contain `label` and `options` properties, simple option objects which contain `value`, `displayName` properties and an optional `selected` property, or a simple array of strings.

_Note: There is an `indentation` optional property if you'd like to "indent" nested `<optgroup>`s. Also note that `<optgroup>`s can't really be nested, hence the need to simulate it using indentation. To disable set indentation to an empty string._

``` js
[{
  label:'Foods',
  options:[
    { value: 'pizza', displayName: "Tasty Pizza"},
    { value: 'waffles', displayName: "Delicious Waffles"},
    { value: 'felafel', displayName: "I'm feelin Falafel"}
  ]
},
{ value: 'react', displayName: "React", selected:true},
{ value: 'angular', displayName: "AngularJS"},
{ value: 'bose', displayName: "No highs no lows, must be Bose"},
{
  label:'States',
  options:[
    'California','Oregon','New York'
  ]
},
{
  label:'Sports',
  options:[{
    label:'Basketball',
    options:['Trail Blazers','Kings']
  },{
    label:'Football',
    options:['49ers','Raiders']
  }]
}
]
```

### DIFF
--- a/src/Dropdown/Dropdown.jsx
+++ b/src/Dropdown/Dropdown.jsx
@@ -20,122 +20,101 @@ const Dropdown = React.createClass({
       React.PropTypes.oneOfType([
         React.PropTypes.string, React.PropTypes.object
       ])).isRequired,
-    initialSelectedInd: React.PropTypes.number,
-    selectedInd: React.PropTypes.number,
-    defaultDisplay: React.PropTypes.string,
     handleChange: React.PropTypes.func.isRequired
   },
 
   getInitialState() {
-    return {open: false};
+    return {value:undefined};
   },
 
   getDefaultProps() {
-    return {data: []};
-  },
-
-  componentDidMount() {
-    document.addEventListener('click', this._checkClickAway);
-  },
-
-  componentWillUnmount() {
-    document.removeEventListener('click', this._checkClickAway);
-  },
-
-  _checkClickAway(e) {
-    if (!ReactDOM.findDOMNode(this.dropdownButton).contains(e.target)) {
-      this.setState({open: false});
-    }
+    return {
+      data: [],
+      label:'',
+      indentation:' • '
+    };
   },
 
   _generateNodes() {
     let {data} = this.props;
+    let {indentation} = this.props;
 
-    // Translate `data` from an array of strings, if necessary.
-    data = data.map(item => {
-      return {
-        value: _.has(item, 'value') ? item.value : item,
-        displayName: _.has(item, 'displayName') ? item.displayName : item,
-        className: item.className,
+    return data.map((item, ind) => (
+      makeOptions(item,ind,0)
+    ));
+
+    function makeOptions(item,ind,depth) {
+      let items = [];
+
+      if(item.options !== undefined) {
+        var optionsDOM = item.options.map((item, ind) => (
+          makeOptions(item,ind,depth + 1)
+        ));
+        var indent = '';
+        for(var i = 0; i < depth; i++) indent += indentation;
+        return (
+          <optgroup label={indent + item.label}>
+            {optionsDOM}
+          </optgroup>
+        );
       }
-    });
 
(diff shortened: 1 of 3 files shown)


src/Dropdown/Dropdown.jsx
```diff
@@ -20,122 +20,101 @@ const Dropdown = React.createClass({
...
+        );
+      }
 
-    // Translate `data` from an array of strings, if necessary.
-    data = data.map(item => {
-      return {
-        value: _.has(item, 'value') ? item.value : item,
-        displayName: _.has(item, 'displayName') ? item.displayName : item,
-        className: item.className,
+      if(typeof(item) !== 'object') { // it is a string
```
aric87: not sure the comments here are correct. typeof [] == 'object' 
also, the comments suggest the last else is for handling objects, but the initial if would handle objects, no?

jpdevries: it is either going to be a typeof `"object"` or `"string"`. If it is not an `"object"` it is a `"string"`.

- if string
    - push created item in
- if not string 
   - if an Array loop and push
   - if not an Array just push

tholex:
Hey JP! I'll pull this version into some apps and try it out this weekend & do a review. Thanks!

jpdevries:
Hey @tholex. How did the test go?

jpdevries:
If for some reason we are against native select elements we still need to do due diligence and add the appropriate ARIA attributed to make browsers and assistive technology treat the unsemantic divs as if they were the native elements they represent. Would also need to script all the other awesome features native input elements come with for free like keyboard shortcuts.

tholex:
Yeah native might work... sorry about sleeping on this. I added it to our trello, will look into it in about a week. It takes a couple hours to set up and QA a change like this.

jpdevries:
Another issue going native would fix, is the lack of implicit accessibility we currently have even for sighted users. Example, how can one tell the difference between any of these options?



The options are too long and we can't tell the difference between them, even if hovering over one of them a tooltip is not displayed. This is one of those interesting situations where you actually have to make accessibility considerations for abled users (CSS and layout generally doesn't affect the accessibility tree)

tholex:
A title tag could (or some other tooltip) could fix that, as would be
putting version first.

JP, I'll try to make the switch but I can't use your code as is since it
drastically changes the API of the thing. Will be working on it in the next
couple of days...

tholex:
Hey JP, I made a version here: https://github.com/Thinkful/ui/pull/356

TLDR: the api changes you made, while they expose neat-o HTML features, are not needed by our apps. It would take a while to edit all ~3 consumer apps and do staged rollouts, so I made a version that hopefully improves accessibility while retaining the same API.

jpdevries:
> A title attr could (or some other tooltip) could fix that, as would be
putting version first.

Well `title` isn't a get out of jail free card. In fact, most a11y experts advocate not using it at all because it is ableist in nature. It has no exposure to the accessibility tree. I think it is fine to use it, as long as you are aware you may need to also set something like `aria-label`. 

> Hey JP, I made a version here: #356

Woohoo! 🎉 

I can't remember what neat-o HTML features might be missing but the main goal was switching to `<select>`.

Issue #316, opened 2016-06-10
Topic Picker Accessibility Needs Label


jpdevries:
Controls and media elements should have labels. Labels can be accessibly hidden if need be.

![](http://f.cl.ly/items/3h1t1j331s092J351q2D/Screen%20Shot%202016-06-10%20at%2010.05.02%20AM.png)

PR #317, opened 2016-06-10
Adding label to TopicPicker

jpdevries:
Controls and media elements should have labels. Labels can be accessibly hidden if need be.

### After

Closes Thinkful/ui#316


jpdevries:
If we aren't going to add a visual `<label>` need to at least use a `.visually-hidden` one or `aria-label`, `aria-labelleby`. The updated design with a visual label seems like it would pass user testing better, as it is a quicker visual read.

Issue #320, opened 2016-06-10
Links (Icons) aren't Accessible


jpdevries:
The purpose of each link should be clear from the link text

![](http://j4p.us/0c0s140J1c3n/Screen%20Shot%202016-06-10%20at%2012.33.33%20PM.png)

PR #321, opened 2016-06-10
Screen Reader Friendly Icons

jpdevries:
I think a full switch to `<svg>` icons would still be a good move but
this makes even our fontface implementations a11y
added `<Icon alt>`

The purpose of each link should be clear from the link text

Closes Thinkful/ui#320


jpdevries:
Since we use icons critically in some cases this is important for the accessibility tree. Currently low vision users will not be able to access various components of or UI.

Issue #322, opened 2016-06-10
OpenSession Overview Bio Photo not Accessible


jpdevries:
Meaningful images should not be used in element backgrounds

![](http://j4p.us/293R1Z28402a/Screen%20Shot%202016-06-10%20at%2012.51.59%20PM.png)

PR #323, opened 2016-06-10
OpenSession Overview Bio Photo a11y

jpdevries:
Meaningful images should not be used in element backgrounds

Closes Thinkful/ui#322


jpdevries:
If I set isWorkshop to false in DemoPage.jsx there is a fatal error, so I haven't tested in QA mode yet. Don't think this one is ready for merge yet.

PR #325, opened 2016-06-17
Upgrade react, router, et al. Bump version

ripleyaffect:
Major version bump (3.0.0)
- Brings all our react up to date
- Brings react-router up to date
- Brings webpack and babel up to date

upgrading apps to this release should be relatively straightforward, as updating from react 0.14 -> 15.0 is pretty painless


client/DemoPage.jsx
```diff
@@ -23,7 +22,7 @@ class DemoPage extends React.Component {
     const sampleDropdownData = [
       { value: 'pizza', displayName: "Tasty Pizza"},
       { value: 'waffles', displayName: "Delicious Waffles"},
-      { value: 'felafel', displayName: "I'm feelin Falafel"},
+      { value: 'falafel', displayName: "I'm feelin Falafel"},
```
rkuykendall: 😱

src/AvailabilityGrid/AvailabilityGrid.jsx
```diff
@@ -82,10 +82,11 @@ const AvailabilityGridDay = React.createClass({
   },
 
   render() {
-    let slotNodes = this.props.data.slots.map((slotData) => {
+    let slotNodes = this.props.data.slots.map((slotData, index) => {
       const {data, ...other} = this.props;
       return (
         <AvailabilityGridSlot
+            key={index}
```
rkuykendall: 👍

rkuykendall:
👍  Looks good. I'm not 100% in love with having to reorder the params to make sure one key overrides another, but it works.

learykara:
I can't get the demo page to run locally... seeing error `"Cannot define 'query' and multiple loaders in loaders list"`... looks related to [this change](https://github.com/Thinkful/ui/pull/325/files#diff-d62dd7a83e030f3c08cfde5b95ad0d80R30) ?

client/DemoPage.jsx
```diff
@@ -99,7 +99,7 @@ class DemoPage extends React.Component {
               data: Demo.props.constant(sampleDropdownData),
               initialSelectedInd: Demo.props.choices([undefined,0,1,2]),
               defaultDisplay: Demo.props.string("Choose something awesome"),
-              handleChange: Demo.props.callback.log(e => e.target.getAttribute('value'))
+              handleChange: Demo.props.callback.log(e => e.target.value)
```
learykara: does this work? `value` is on a `p` tag in the dropdown - thought it needed to be accessed using `getAttribute` but I could be wrong

ripleyaffect: Yes, but it's one of the things that needs to be checked for when an app upgrades. You should be able to run the demo now and confirm

ripleyaffect:
I've removed eslint as a loader after attempting to use it as a preloader (it was causing the above issue with query params on multiple loaders), as it was preventing the hotload server from running correctly. We should revisit linting elsewhere (maybe as a npm script)

tholex:
👍 god's work. Maybe let's keep the eslintrc, just not the loader? +1 on doing the actual linting in npm scripts too

jpdevries:
Very nice. Looks like this bumps to 3.0, thus making breaking changes. 
I've submitted a few PRs that make incremental accessibility and usability improvements without breaking changes. If we do a major bump I would +1 a more agressive accessibility audit. 
Not to be dramatic, but many of our components currently don't respect accessibility laws. I think this is something we should address.

rkuykendall:
haha, just got a Github notification about this. :wave: Hi Thomas!

PR #327, opened 2016-06-22
Check for impersonation cookie.

rkuykendall:


src/analytics/actions.es6
```diff
@@ -32,13 +32,15 @@ function isLoggedIn() {
 }
 
 function isImpersonating() {
-    const impersonating = has(global, '__env.user.real_admin_tf_login');
+    const inEnv = has(global, '__env.user.real_admin_tf_login');
+    const inCookie = has(cookies, 'Tf-Impersonating');
 
-    if (impersonating) {
+    if (inEnv || inCookie) {
         log('No analyitcs for impersonating users.');
```
learykara: `analyitcs`

tholex:
👍 as long as `Tf-Impersonating` is the right string / cookie name, not sure where to double check that

tati:
LGTM 👍

src/analytics/actions.es6
rkuykendall: 👍

rkuykendall:
@tholex Double checked here https://github.com/Thinkful/hawk/pull/191 👍

PR #331, opened 2016-06-23
Block segment from loading at all if user is impersonating.

rkuykendall:
If this doesn't work, we're hopeless.


tati:
😂  😂  😂  👍

PR #333, opened 2016-06-29
Update getstream to 3.4

ripleyaffect:
[faye 1.2.0](https://github.com/faye/faye/releases/tag/1.2.0) appears to break semver (though unsure how). The upgrade from `getstream` 2.x to 3.x looks like an in-place upgrade for us.

We use the following methods:
- `stream.connect` (option with api key and app key)
- `client.feed`
- `feed.get` (with options `limit`, `mark_seen`, and `mark_read`)
all of which remain the same between 2.x and 3.x

`getstream` 3.4 pins faye at 1.2.0

[:metal:🎶](https://www.youtube.com/watch?v=LYU-8IFcDPw) I can't faye...the way I did before! Don't break semver on me! I won't be ignored! [🎶:metal:](https://www.youtube.com/watch?v=LYU-8IFcDPw)

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "eslint-loader": "^0.12.0",
     "extract-text-webpack-plugin": "^0.8.1",
     "file-loader": "^0.8.4",
-    "getstream": "^2.2.2",
+    "getstream": "^3.4.0",
     "history": "^1.13.1",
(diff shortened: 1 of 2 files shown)


diezguerra:
👌

PR #334, opened 2016-07-20
Feature/update in app footer

learykara:
[This trello card](https://trello.com/c/hs2AYUig/499-update-in-app-footer) and CP name change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "thinkful-ui",
-  "version": "2.0.79",
+  "version": "2.0.80",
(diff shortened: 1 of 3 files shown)


tholex:
If we'll be supporting the new URL's maybe change the URL's too? (e.g. from `/courses/web-development-career-path/` to whatever is next)

Also "Career Path" in here:
https://github.com/Thinkful/ui/blob/master/src/AppBar/Navigation.jsx

👍 otherwise

learykara:
....yeah. definitely meant to change the URLs too. thanks for catching that. there are too many things going on today.

PR #335, opened 2016-08-08
remove is_js import

learykara:


src/analytics/actions.es6
```diff
@@ -15,6 +14,15 @@ const cookies = zipObject(map(document.cookie.split('; '), function(cookie) {
     return [name, decodeURIComponent(value)];
 }));
 
+function is(thing, type) {
+    return typeof(thing) === type
```
ripleyaffect: hmm, this is a little sketchy. for example `typeof([]) === 'object'`. `Object.prototype.toString.call` is better to use:
```es6
function is(thing, type) {
  switch(type) {
    case 'array':
      return Object.prototype.toString.call(thing) === '[object Array]'
    default:
      throw new Error(`Type "${type}" not recognized`)
  }
}

```

ripleyaffect:
:+1: apart from comment

PR #341, opened 2016-09-29
update logged out nav to match pelican

learykara:
The line count may look scary but it's basically a copy + paste from pelican with reactified interactions for dropdowns, etc.

screen shot 2016-09-29 at 11 08 19 am


jcudmore:
👍 !!

PR #344, opened 2017-01-24
add profile function

tati:
Adds a new function that should only update or add a new profile with the passed data as opposed to also changing the distinct id of the profile.

If you use segment's `identify` function without passing an id it will not change the distinct id of the profile (which is what I am trying to do here).

This is not be confused with mixpanel's identify function which is something different.

For reference
[We have an identify function](https://github.com/Thinkful/ui/blob/master/src/analytics/actions.es6#L217)
[Segment has one](https://github.com/segment-integrations/analytics.js-integration-mixpanel/blob/master/lib/index.js#L156)
[Mixpanel has one](https://mixpanel.com/help/reference/javascript-full-api-reference#mixpanel.identify)
and they all do slightly different things depending on what you pass into them :woman_facepalming:


src/analytics/actions.es6
```diff
@@ -245,6 +245,25 @@ function identify(id, traits, options, fn) {
         global.analytics.identify(id, traits, options, fn);
 }
 
+function profile(traits) {
```
tati: how do you feel about 1) the name 2) I know this is a lot of comments that could get out of date but I WISH someone had left more comments for me on this

tholex: How does it know which profile to update? Do you have to pass an `id` in this object?

tholex: This is missing a lot of features of identify, maybe just detect a call like `identify(null, traits..)` and call `global.analytics(traits...)` ?

src/analytics/actions.es6
```diff
@@ -245,6 +245,25 @@ function identify(id, traits, options, fn) {
...
+  // creates a new profiles or updates a profile with new data
+
+  if (typeof(traits) != "object") {
+    console.error("You must pass an object like {email: 'hello@thinkful.com'}");
+    return
+  }
```
tholex: also indentation

src/analytics/actions.es6
```diff
@@ -245,6 +245,25 @@ function identify(id, traits, options, fn) {
...
+
+    if (isImpersonating()) {
+      // no need to update data if someone is just impersonating
+      return;
+    }
+
+    // "identify" is segment.io's function not mixpanel's
+    // see https://github.com/segment-integrations/analytics.js-integration-mixpanel/blob/master/lib/index.js
```
tholex: No need for these comments.

PR #346, opened 2017-01-26
make emails lowercase

tati:
Somewhere emails are dupping with caps profiles...trying to pin down where but no getting the issue locally

Ex https://github.com/Thinkful/Thinkful/issues/849

### DIFF
--- a/src/analytics/actions.es6
+++ b/src/analytics/actions.es6
@@ -71,7 +71,7 @@ function isImpersonating() {
 function tryEmail() {
     // Check if the user is logged in
     if (has(global, '__env.user.tf_login')) {
-        return global.__env.user.tf_login;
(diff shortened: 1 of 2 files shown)


src/analytics/actions.es6
```diff
@@ -236,7 +236,7 @@ function identify(id, traits, options, fn) {
 
     // If someone is passing in an email, let's assign it to the traits
     if (id && isEmail(id)) {
-        traits.email = id;
+        traits.email = id.toLowerCase();
```
tholex: maybe just lowercase the `id` param earlier? you're using `id` later without lowercasing it

tati: if some cases the id isn't an email though and in that case I don't want to lowercase it...just need it lowercased inside this if block cuz then it's an email

PR #347, opened 2017-02-07
date picker improvements

learykara:
Some improvements to the `DatePicker` -
- defaults to "today" as the active day if none passed
- dropdown closes when you click somewhere else on the screen
- supports displaying a placeholder value in the dropdown until a date is picked (such as "Select a date", etc)
- use thinkful ui `Icon` component

### DIFF
--- a/src/DatePicker/DatePicker.jsx
+++ b/src/DatePicker/DatePicker.jsx
@@ -111,12 +111,10 @@ class DatePicker extends React.Component {
   }
 
   _checkClickAway(event) {
-    const buttonDomNode = ReactDOM.findDOMNode(this.dropdownButton)
-    const calendarDomNode = ReactDOM.findDOMNode(this.calendar)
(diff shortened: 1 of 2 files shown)


src/DatePicker/DatePicker.jsx
```diff
@@ -97,12 +110,26 @@ class DatePicker extends React.Component {
     });
   }
 
+  _checkClickAway(event) {
+    const buttonDomNode = ReactDOM.findDOMNode(this.dropdownButton)
+    const calendarDomNode = ReactDOM.findDOMNode(this.calendar)
```
ripleyaffect: you shouldn't need to call `findDOMNode` here, since the components are html elements already

diezguerra:
@learykara forgot to remove the react-dom import :inquisition:

learykara:
Oh no the worst has happened!!! Fortunately I had not yet published.
Removed.

PR #349, opened 2017-02-28
Link to avaialble-students

ripleyaffect:
Link to the new available-students screen from the `linkSet`


src/AppBar/linkSet.es6
```diff
@@ -46,7 +46,11 @@ else {
         main.push(home);
 
         main.push(config.qaSessions);
-        menu.push(config.takeStudent);
+        menu.push({
+            displayName: 'Available Students',
```
learykara: are we title casers now? I can never keep track

PR #351, opened 2017-03-11
fix(analytics): remove tailorbird, consider `Stork`

diezguerra:


diezguerra:
it is https://github.com/Thinkful/ui/blob/435aaff45961500329acdd68b3a5439b8dc2e8db/src/analytics/actions.es6#L42

PR #356, opened 2017-03-20
HTML-ify the Dropdown component

tholex:
Improves usability / accessibility of the dropdown component.

Retains existing API for the component, adds label-friendly `id` option.

JP's original edition: https://github.com/Thinkful/ui/pull/315

### DIFF
--- a/src/Dropdown/Dropdown.jsx
+++ b/src/Dropdown/Dropdown.jsx
@@ -34,20 +34,6 @@ const Dropdown = React.createClass({
     return {data: []};
   },
 
-  componentDidMount() {
-    document.addEventListener('click', this._checkClickAway);
-  },
-
-  componentWillUnmount() {
-    document.removeEventListener('click', this._checkClickAway);
-  },
-
-  _checkClickAway(e) {
-    if (!ReactDOM.findDOMNode(this.dropdownButton).contains(e.target)) {
-      this.setState({open: false});
-    }
-  },
-
   _generateNodes() {
     let {data} = this.props;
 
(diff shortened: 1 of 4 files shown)


src/Dropdown/Dropdown.jsx
```diff
@@ -99,44 +82,25 @@ const Dropdown = React.createClass({
...
+      "tui-dropdown-container",
+      dropdownClasses);
+
+    const selectedInd = this._determineSelectedInd();
+
+    return <div className={dropdownClasses}>
+      <select
+        id={htmlId}
+        onChange={e => this._handleChange(e)}
+        className="tui-dropdown">
```
ripleyaffect: needs an extra indent to visually distinguish from the children

src/Dropdown/Dropdown.jsx
```diff
@@ -62,24 +51,18 @@ const Dropdown = React.createClass({
 
     return data.map((item, ind) => {
       return (
-        <p
-          className={cx("dropdown-item", item.className)}
-          id={ind}
+        <option
+          className={cx(item.className)}
```
ripleyaffect: no need to call `cx` here, unless you're supporting `item.className` as not a string

src/Dropdown/Dropdown.jsx
```diff
@@ -99,44 +82,25 @@ const Dropdown = React.createClass({
...
+
+    const selectedInd = this._determineSelectedInd();
+
+    return <div className={dropdownClasses}>
+      <select
+        id={htmlId}
+        onChange={e => this._handleChange(e)}
+        className="tui-dropdown">
+        {selectedInd == -1 &&
+          <option selected disabled key="tui-dropdown-default">{defaultDisplay}</option>}
```
ripleyaffect: `defaultDisplay` isn't required and there's no default, should have one of those, or this should default to having the first option selected

src/Dropdown/Dropdown.jsx
tholex: done

src/Dropdown/Dropdown.jsx
tholex: done.

src/Dropdown/Dropdown.jsx
tholex: HTML's default behavior selects the first element, it works without adding the `selected` attribute.

tholex: http://codepen.io/tholex/pen/99c3ab917e0c275f5dde4fca58519d1d

PR #357, opened 2017-03-21
Hotfix/dropdown issue

tholex:
Fixes dropdown issue which was causing cuckoo errors earlier...

This doesn't look right (`selected` being not used), but React recommends this method – http://stackoverflow.com/a/24574465/126963

### DIFF
--- a/src/Dropdown/Dropdown.jsx
+++ b/src/Dropdown/Dropdown.jsx
@@ -39,7 +39,6 @@ const Dropdown = React.createClass({
 
   _generateNodes() {
     let {data} = this.props;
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.jsx
```diff
@@ -39,13 +39,14 @@ const Dropdown = React.createClass({
 
   _generateNodes() {
     let {data} = this.props;
+    const selectedInd = this._determineSelectedInd();
```
learykara: is this used?

tholex: nope! thanks 👍

PR #360, opened 2017-04-04
feat(analytics): undo mixpanel workarounds

diezguerra:
No need to tip-toe around identifying and aliasing anymore. Go ham.

### DIFF
--- a/src/analytics/actions.es6
+++ b/src/analytics/actions.es6
@@ -97,19 +97,8 @@ function getUserId(id) {
         return global.__env.user.tf_login;
     }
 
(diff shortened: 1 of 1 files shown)


src/analytics/actions.es6
```diff
@@ -280,7 +248,7 @@ function alias(to, from, options, fn) {
 
       // Create a track event so we can see exactly when a user is aliased,
       // to help us better understand everything.
-      track('aliased', { 'to': to })
+      track('aliased', {'from': from, 'to': to })
```
ripleyaffect: space before 'from', also no need for string keys.

PR #363, opened 2017-06-13
Fix typo and add link to student activity

Carlosedo:
Fixes a typo that was breaking open-sessions page:

image

Since I was editing `ui` I made another improvement to my previous changes

### DIFF
--- a/src/OpenSessionOverview/OpenSessionOverview.jsx
+++ b/src/OpenSessionOverview/OpenSessionOverview.jsx
@@ -81,14 +81,18 @@ class OpenSessionOverview extends React.Component {
   }
 
   _renderGravatars() {
-    return this.props.participants.map((p, idx) => {
+    const {config, participants} = this.props;
+
+    return participants.map((p, idx) => {
       return (
         <div className="qa-session-participants__item" key={idx}>
-          <Gravatar
-            className="gravatar-image__participant"
-            src={p.image_url}
-            email={p.email}/>
-          <div>{p.name}</div>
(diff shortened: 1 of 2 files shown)


src/OpenSessionOverview/OpenSessionOverview.jsx
```diff
@@ -81,14 +81,18 @@ class OpenSessionOverview extends React.Component {
...
+
+    return participants.map((p, idx) => {
       return (
         <div className="qa-session-participants__item" key={idx}>
-          <Gravatar
-            className="gravatar-image__participant"
-            src={p.image_url}
-            email={p.email}/>
-          <div>{p.name}</div>
+          <a href={`${config.activity.url}/student/${p.user_id}`}>
```
learykara: this isn't linking to myna is it? I think should go to their dashboard activity url if so

Carlosedo: It is. I did the same we do here: https://github.com/Thinkful/cuckoo/blob/master/cuckoo/client/components/WorkshopDetailPage/WorkshopDetailPage.jsx#L155

I'll change it to link to the dashboard. Should we make the change in cuckoo too?

learykara: that'd be 💯 if you want to do it now, if not I'd say to open a cleanup card for it

Carlosedo: 👍

PR #365, opened 2017-08-29
disable rsvp button on click

Carlosedo:
Solves: https://bloopers.thinkful.com/thinkful/pysplash/issues/267/events/latest/

Long time without using React so 🙏, double check haha

### DIFF
--- a/src/OpenSessionOverview/OpenSessionOverview.jsx
+++ b/src/OpenSessionOverview/OpenSessionOverview.jsx
@@ -34,6 +34,9 @@ class OverviewContent extends React.Component {
 class OpenSessionOverview extends React.Component {
   constructor(props) {
     super(props);
+    this.state = {
+      requestPending: false
+    }
     this._renderQASessionCtas = this._renderQASessionCtas.bind(this);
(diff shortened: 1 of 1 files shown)


src/OpenSessionOverview/OpenSessionOverview.jsx
```diff
@@ -98,6 +99,11 @@ class OpenSessionOverview extends React.Component {
     });
   }
 
+  _handleRSVPButtonClick(e, id) {
+    $(e.target).addClass('button__disabled');
+    handleRSVPClick(id);
```
learykara: I'd actually do this via the state.  Something like `this.setState({ requestPending: true })`, and use that on the component to add & remove the disabled class

Carlosedo:
updated it @learykara 
please take another look 😄

learykara:
what happens when the request is finished? does the component/page reload or does it stay in a loading state forever?

Carlosedo:
When the request is finished `attending` is set to true and the button is hidden
https://github.com/Thinkful/ui/blob/master/src/OpenSessionOverview/OpenSessionOverview.jsx#L290

PR #369, opened 2018-01-11
Move styles into lib

ripleyaffect:
- Move all `.less` files into a new `lib` directory

This is the first step of a migration away from an unbuilt ui package.

The `lib` directory will become the `main` of the package, and the webpack config will be removed before the release is generally available.

The javascript will be built with babel's cli and consumers will need to import the less directly from `thinkful-ui/styles.less` or from the appropriate style module if they're size-conscious


package-lock.json
```diff
@@ -0,0 +1,6996 @@
+{
```
learykara: adding this everywhere is a thing we're doing now? I remember carlos sharing something that said it should be done in shoebill, just want to be sure

build/main.css
```diff
@@ -0,0 +1,2 @@
...it-transform:translate(-50%,0);transform:translate(-50%,0)}.tui-notification-list:after{border-left:6px solid transparent;border-right:6px solid transparent;border-bottom:10px solid #FFF;-webkit-transform:translate(-50%,0);transform:translate(-50%,0)}.tui-notification-item{z-index:5;position:relative;list-style:none;margin-left:0;margin-bottom:0;padding:9px 15px 10px;width:330px;max-width:100%;background-color:#FFF}.tui-notification-item+.tui-notification-item{border-top:1px solid #EEEEEF}.tui-notification-item:hover{background:#FAFAFA}.tui-notification-item-dismiss{position:absolute;top:0;right:0;width:30px;height:30px;line-height:30px;text-align:center;color:#8E939A}.tui-notification-item-dismiss:hover{color:#404853}.tui-notification-item-dismiss .icon-close{font-size:24px}.tui-notification-content{display:block;color:#1E2835}.tui-notification-content:hover{color:#1E2835}.tui-notification-time{display:block;font-size:12px;line-height:20px;color:#8E939A;text-transform:uppercase}.tui-notification-message{margin-bottom:0;line-height:20px;font-size:14.5px}.tui-notification-voting-bar{direction:rtl;padding-top:10px;text-align:center}.tui-voting-star+.tui-voting-star{padding-right:5px}
```
learykara: so in other repos, what's in `build/` is generated in ci before pushing. here we're going to do taht locally and commit build/* when changes are made?

package-lock.json
ripleyaffect: yeah, this should be committed for npm5

PR #391, opened 2018-07-31
update Support to Student success

learykara:
rename `Support` -> `Student success`


learykara:
@chenatlas turns out this change shouldn't happen after all

PR #393, opened 2018-11-20
remove unused components, update styles to use sg v 6

learykara:
this line diff isn't nearly as bonkers as it looks.  it's mostly deleting unused components/styles, updating color variable names, and package-lock.json


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "thinkful-ui",
-  "version": "4.0.0-alpha.12",
-  "description": "Shared navigation and UI resources for Thinkful.",
+  "version": "5.0.0",
```
ripleyaffect: this should probably just be updated to match the version of styleguide. it's a pattern used by libs that have releases happen in lockstep (e.g. react)

stories/index.stories.js
```diff
@@ -39,3 +39,38 @@ storiesOf('Headers', module)
...
+  .add('Basic', () => <Footer config={env.config} />)
+
+storiesOf('Icon', module)
+  .add('Navigate right', () => <Icon name="navigateright" />)
+
+storiesOf('Loader', module)
+  .add('Basic', () => <Loader />)
+
+storiesOf('Modal', module)
+  .add('Basic', () => <Modal onClose={() => {}}><h2>Hello, world</h2></Modal>)
```
ripleyaffect: :wave:

PR #397, opened 2018-11-21
add Avatar component

learykara:
image


src/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,52 @@
...
+  contactId: PropTypes.number,
+  email: PropTypes.string,
+  imageUrl: PropTypes.string,
+  size: PropTypes.number,
+};
+
+Avatar.defaultProps = {
+  config: null,
+  email: '',
+  size: 48,
```
ripleyaffect: might as well move border radius here instead and not bother w/ the less

src/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+class Avatar extends React.Component {
+  getImageUrl() {
+    const { contactId, email, imageUrl } = this.props;
+    const config = this.props.config || __env.config;
```
ripleyaffect: should use `global.__env` and can make this the default `config` prop

src/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+Avatar.propTypes = {
+  config: PropTypes.shape({
+    api: PropTypes.shape({
+      url: PropTypes.string,
+    }),
+  }),
+  contactId: PropTypes.number,
+  email: PropTypes.string,
+  imageUrl: PropTypes.string,
```
ripleyaffect: maybe there is a default one? like the TF logo or something?

PR #398, opened 2018-11-22
Hotfix/topic picker tag

learykara:
lotsa jsprettier changes in here taht can be ignored


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "thinkful-ui",
-  "version": "6.0.0",
+  "version": "6.0.1",
```
ripleyaffect: can be 2 now!

PR #400, opened 2018-11-24
use background image to prevent distortion

learykara:


src/Avatar/Avatar.jsx
```diff
@@ -35,10 +35,15 @@ class Avatar extends React.Component {
...
     return (
-      <img
+      <div
         className={cx('user-avatar', className)}
-        style={{ borderRadius: '2px', width: `${size}px`, height: `${size}px` }}
-        src={this.getImageUrl()}
+        style={{
+          background: `url(${this.getImageUrl()})`,
+          backgroundSize: 'cover',
+          borderRadius: '2px',
```
ripleyaffect: I think It may make sense to keep this in terms of a ratio as well, so the style stays consistent when the size is scaled

src/Avatar/Avatar.jsx
```diff
@@ -35,10 +35,15 @@ class Avatar extends React.Component {
     } = this.props;
 
     return (
-      <img
+      <div
         className={cx('user-avatar', className)}
-        style={{ borderRadius: '2px', width: `${size}px`, height: `${size}px` }}
-        src={this.getImageUrl()}
+        style={{
+          background: `url(${this.getImageUrl()})`,
```
ripleyaffect: could be `url(${this.getImageUrl()}) no-repeat center center fixed`

PR #402, opened 2018-11-29
Feature/rework link set

learykara:


src/AppBar/Navigation.jsx
```diff
@@ -55,11 +54,15 @@ class AppNav extends React.Component {
...
   }
+
   renderAuthed(user, config) {
+    const { linkSet } = this.state;
+
     const navClassName = cx('app-nav', {
       'app-nav__visible': this.state.isMenuVisible,
     });
-    const navLinks = this.linkSet.main.filter(link => !link.search);
+    const navLinks = linkSet.main.filter(link => !link.search);
```
ripleyaffect: can `search` just not be added in the link set?

src/AppBar/linkSet.js
```diff
@@ -1,82 +1,55 @@
...
+    menu.push({
+      displayName: 'Available Students',
+      host: brandConfig.lark.host,
+      url: brandConfig.lark.url + '/available-students/',
+    });
+  }
+
+  // TF-brand students only
+  if (user.brand === 'thinkful') {
+    menu.push(brandConfig.refer);
```
ripleyaffect: I think this may be better as a function that adds brand-specific links. Actually this whole function is probably better as a set of getters composed into a `[].concat`:

```es6
const menu = [].concat(
  getDefaultLinks(...),
  getUserLinks(...),
  getBrandLinks(...),
  getCommonLinks(...)
)
```

Though better to do that piece by piece, not all in this PR

PR #405, opened 2018-11-30
style desktop nav

learykara:
image


src/AppBar/Navigation.jsx
```diff
@@ -91,6 +94,10 @@ class AppNav extends React.Component {
                   </div>
                 )}
               </a>
+              {/* Desktop toggle */}
+              <a className="app-nav-arrow" onClick={this._toggleMenu}>
+                <Icon name="navigatedown" />
+              </a>
```
ripleyaffect: these toggles are also good examples of components to split out

PR #409, opened 2018-11-30
add SvgIcon and book

learykara:
only adding one icon for now but I'll get em all over here 🔜


src/SvgIcon/Icons/Book.jsx
```diff
@@ -0,0 +1,34 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import cx from 'classnames';
+
+import sgVars from '../../styleguide';
```
ripleyaffect: why not `{ darkBlue }`?

src/SvgIcon/Icons/Book.jsx
```diff
@@ -0,0 +1,34 @@
...
+Book.propTypes = {
+  color: PropTypes.string,
+  size: PropTypes.number.isRequired,
+};
+
+Book.defaultProps = {
+  color: sgVars.darkBlue,
+};
+
+module.exports = Book;
```
ripleyaffect: you can `export default` here, since you're using `import` in `SvgIcon`. exports/imports just need to be moved over to this style. It can wait until after your navbar work though

PR #412, opened 2018-11-30
prefix class names, add UnauthedAppBar

learykara:
image


src/AppBar/UnauthedAppBar.jsx
```diff
@@ -0,0 +1,34 @@
...
+            </ul>
+          </div>
+        </div>
+      </nav>
+    </div>
+  );
+};
+
+UnauthedAppBar.propTypes = {
+  config: PropTypes.object.isRequired,
```
ripleyaffect: shape w/ www and accounts?

PR #421, opened 2018-12-05
support enrollment switcher in app bar

learykara:


package.json
```diff
@@ -75,7 +77,7 @@
     "spark-md5": "0.0.5",
     "superagent": "^1.4.0",
     "superagent-no-cache": "^0.1.0",
-    "tfstyleguide": "^6.1.0"
+    "tfstyleguide": "^6.1.2"
```
ripleyaffect: two hops this time! CRISS CROSSSS

src/AppBar/AppBar.jsx
```diff
@@ -78,6 +78,7 @@ class AppBar extends React.Component {
               <a className="tui-app-nav-logo" href={linkSet.home.url}>
                 <Logo brand={user.brand} />
               </a>
+              {EnrollmentView && <EnrollmentView />}
```
ripleyaffect: is this not just `{EnrollmentView}` to support the extra class on the one below? This is a bit of a weird slot pattern, typically you'd pass `<EnrollmentView />` directly into the slot

PR #422, opened 2018-12-05
switch footer attrs by brand

learykara:


src/Footer/Footer.jsx
```diff
@@ -3,9 +3,11 @@ import React from 'react';
 import cx from 'classnames';
 import moment from 'moment';
 
-const Footer = ({ className, config, user }) => {
+const Footer = ({ className, config, user, brand }) => {
   // Can't be set via defaultProps because of frontend testing and global.__env
   config = config || global.__env.config;
+  const displayBrand = brand || user.brand || 'thinkful';
+  const brandConfig = _.assign({}, config, config.brands[displayBrand] || {});
```
ripleyaffect: you don't need the object fallback here. assigning undefined does nothing

PR #427, opened 2019-01-15
supress call to stream API for DS users

learykara:
We're reaching our API limit with stream, which we use for the nav bar notifications.  Design system users don't get those notifications, we should suppress any calls to the API for them

### DIFF
--- a/src/AppBar/notifications/NotificationActions.jsx
+++ b/src/AppBar/notifications/NotificationActions.jsx
@@ -1,17 +1,34 @@
 const Reflux = require('reflux');
 const stream = require('getstream');
 const _ = require('lodash');
+
 const CONFIG = global.__env ? global.__env.config : null;
 const USER = global.__env ? global.__env.user : null;
+const DESIGN_SYS_FLAG = 'design-system';
+const LIMIT = 5;
+
 const NotificationActions = Reflux.createActions({
   fetchNotifications: { asyncResult: true },
   markSeen: { asyncResult: true },
   markRead: { asyncResult: true },
   processEvent: { asyncResult: true },
 });
+
 let client = null;
 let userFeed = null;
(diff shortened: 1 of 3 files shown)


src/AppBar/notifications/NotificationActions.jsx
```diff
@@ -22,9 +39,8 @@ if (CONFIG && CONFIG.vendor.getstream.userFeedToken) {
     (USER ? USER.contact_id : 1).toString(),
     CONFIG.vendor.getstream.userFeedToken,
   );
-} else {
```
ripleyaffect: lol that this was here

src/AppBar/notifications/NotificationActions.jsx
```diff
@@ -1,17 +1,34 @@
...
 let userFeed = null;
-if (CONFIG && CONFIG.vendor.getstream.userFeedToken) {
+
+const shouldInitNotifications = () => {
+  // Initialize notifications if we have the token in config and the user
+  // is not on the design system
+  return (
+    CONFIG &&
+    USER &&
+    CONFIG.vendor.getstream.userFeedToken &&
```
ripleyaffect: This may be not be present even if `CONFIG` is

src/AppBar/notifications/NotificationActions.jsx
```diff
@@ -1,17 +1,34 @@
 const Reflux = require('reflux');
 const stream = require('getstream');
 const _ = require('lodash');
+
 const CONFIG = global.__env ? global.__env.config : null;
 const USER = global.__env ? global.__env.user : null;
+const DESIGN_SYS_FLAG = 'design-system';
+const LIMIT = 5;
```
ripleyaffect: what limit is this? name is too vague (I know it was just moved)

PR #434, opened 2019-06-11
Add curtainColor prop.

seanohue:
This is to support doing the work here:
https://github.com/Thinkful/seagull/pull/218
In a less hack-y way.

### DIFF
--- a/src/Modal/Modal.jsx
+++ b/src/Modal/Modal.jsx
@@ -34,9 +34,9 @@ class Modal extends React.Component {
   }
 
   render() {
-    const { className } = this.props;
+    const { className, curtainColor } = this.props;
     const { controlledByParent, isOpen } = this.state;
-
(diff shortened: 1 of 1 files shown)


src/Modal/Modal.jsx
```diff
@@ -58,7 +58,12 @@ class Modal extends React.Component {
...
 Modal.propTypes = {
   className: PropTypes.string,
-  close: PropTypes.func
-}
+  close: PropTypes.func,
+  curtainColor: PropTypes.string,
+};
+
+Modal.defaultProps = {
+  curtainColor: '',
```
mikebranski: Since `''` is not a valid CSS background color, and we probably don't want to force `transparent` all the time, I think we should conditionally apply the `backgroundColor` style up above based on the presence of this prop.

PR #439, opened 2019-07-25
check for user platform

learykara:


src/AppBar/AppBar.jsx
```diff
@@ -59,10 +61,7 @@ class AppBar extends React.Component {
   _shouldInitNotifications() {
     const { user } = this.props;
 
-    return (
-      !/mentor|admin/.test(user.role) &&
-      user.access.indexOf('design-system') === -1
-    );
+    return !/mentor|admin/.test(user.role) && user.platform === LEGACY_PLATFORM;
```
ripleyaffect: ditto r.e. turning this into an `||` for now

PR #443, opened 2019-11-15
[T12386] Flexperiment: Concierge

alis0nc:
https://phabricator.bloc.io/T12386

This PR adds the Concierge experiment to `AppBar` behind a feature flag. 

### Desktop
Screen Shot 2019-11-19 at 11 35 04
Screen Shot 2019-11-19 at 11 35 28
### Mobile
Screen Shot 2019-11-19 at 11 35 17
Screen Shot 2019-11-19 at 11 35 41


less/modules/app-bar.less
```diff
@@ -82,8 +82,14 @@
   position: relative;
 }
 
-.tui-app-nav-list {
+.visibility-transition {
+  transition: transform 0.35s ease, opacity 0.28s ease, visibility 0.28s ease, max-width 0.35s ease,
+    max-height 0.22s ease;
+}
+
+.tui-app-bar-modal {
```
alis0nc: I pulled these out into a mixin because the hiding/unhiding is common between the nav menu and the concierge

less/modules/app-bar.less
```diff
@@ -140,17 +154,166 @@
...
+  height: 0;
+  border-left: @concierge-arrow-size solid transparent;
+  border-right: @concierge-arrow-size solid transparent;
+  border-bottom: @concierge-arrow-size solid @white;
+  position: absolute;
+  top: calc(@nav-height - @concierge-arrow-size);
+  z-index: 102; // this is +1 than the concierge's because we want the arrow
+                // to be above the box shadow
+  visibility: hidden;
+  right: 69px; // nice
```
seanohue: LOL

less/modules/app-bar.less
```diff
@@ -82,8 +82,14 @@
   position: relative;
 }
 
-.tui-app-nav-list {
+.visibility-transition {
+  transition: transform 0.35s ease, opacity 0.28s ease, visibility 0.28s ease, max-width 0.35s ease,
```
seanohue: Are the animation speeds able to be made into a LESS var for reuse?

less/modules/app-bar.less
```diff
@@ -257,6 +420,42 @@
   }
 }
 
+@concierge-toggle-diameter: 32px;
+
+.tui-app-nav-concierge-toggle {
+  height: @concierge-toggle-diameter;
```
seanohue: Doesn't have to be done now (should wait until we do this in a couple of other spots) but I feel like this could be pulled into a parametric mixin for making a circle in CSS.

alis0nc: ooh good idea (also I was overthinking it and could have set `border-radius: 50%`

less/modules/app-bar.less
alis0nc: IMO moving `.visibility-transition` into a mixin is sufficient

src/AppBar/AppBar.jsx
```diff
@@ -66,8 +86,8 @@ class AppBar extends React.Component {
 
   render() {
     const { brand, className, config, EnrollmentView, user } = this.props;
-    const { isMenuVisible, linkSet } = this.state;
-
+    const { isMenuVisible, isConciergeVisible, linkSet } = this.state;
+
```
ripleyaffect: 😱

src/AppBar/AppBar.jsx
```diff
@@ -97,6 +118,12 @@ class AppBar extends React.Component {
               </ul>
             </div>
             <div className="tui-app-nav-right">
+              {_.includes(user.access, CONCIERGE_FLAG) &&
+                <ConciergeToggle
+                  conciergeVisible={isConciergeVisible}
+                  onClick={this._toggleConcierge}
+                />
+              }
```
ripleyaffect: should be wrapped in parens:
```es6
{_.includes(user.access, CONCIERGE_FLAG) && (
  <ConciergeToggle
    conciergeVisible={isConciergeVisible}
    onClick={this._toggleConcierge}
  />
)}
```

src/AppBar/ConciergeModal.jsx
```diff
@@ -0,0 +1,95 @@
...
+    Beta
+  </span>
+);
+
+const ConciergeOption = ({ title, subtitle, link, background }) => (
+  <a href={link}>
+    <div
+      className="tui-concierge-option"
+      style={{
+        background: background,
```
ripleyaffect: don't need the duplicate, b/c of object key/value expansion: `{ background }`

src/AppBar/ConciergeModal.jsx
```diff
@@ -0,0 +1,95 @@
...
+ConciergeOption.propTypes = {
+  title: PropTypes.string.isRequired,
+  subtitle: PropTypes.string.isRequired,
+  link: PropTypes.string.isRequired,
+  background: PropTypes.string,
+}
+
+ConciergeOption.defaultProps = {
+  background: '#000',
+};
```
ripleyaffect: components should be in their own files, if only to keep the files clean and single-purpose

stories/index.stories.js
```diff
@@ -44,6 +44,14 @@ storiesOf('AppBar', module).add('Logged in', () => (
   <AppBar config={env.config} user={env.user} />
 ));
 
+storiesOf('AppBar', module).add('Concierge experiment', () => {
+  const user = {
+    ...env.user,
+    access: ['flexperiment-concierge'],
```
ripleyaffect: maybe `user.access.concat('flexperiment-concierge')`? just to ensure you're not forcing replacement

less/modules/app-bar.less
seanohue: Agreed that consolidating/library-ifying transitions & animations could be a bigger cleanup card and doesn't need to be done as part of this.

src/AppBar/ConciergeModal.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+ConciergeOption.propTypes = {
+  title: PropTypes.string.isRequired,
+  subtitle: PropTypes.string.isRequired,
+  link: PropTypes.string.isRequired,
+  background: PropTypes.string,
+}
+
+ConciergeOption.defaultProps = {
+  background: '#000',
```
seanohue: I feel like this might already be a var somewhere?

src/AppBar/ConciergeModal.jsx
```diff
@@ -0,0 +1,95 @@
...
+        </p>
+      </div>
+    </div>
+
+    <h3>Stuck? Let's get you some help.</h3>
+    <div className="tui-concierge-options-container">
+      <ConciergeOption
+        title="Start a convo on Slack"
+        subtitle="Pair with other students on the same topics"
+        background="#1733ff"
```
seanohue: This and other colors may also be vars, if nothing else then in [tfstyleguide](https://github.com/Thinkful/tfstyleguide). I feel like there was a helper somewhere that pulled the LESS vars into JavaScriptland. Not really a blocker though esp. if it's hard to find.

src/AppBar/ConciergeModal.jsx
alis0nc: `@black` in tfstyleguide

src/AppBar/ConciergeModal.jsx
seanohue: Found it: https://github.com/Thinkful/ui/tree/master/src/styleguide

stories/index.stories.js
alis0nc: good point -- it's empty in the mock user for stories, but that doesn't mean it will always empty

src/AppBar/ConciergeModal.jsx
alis0nc: The background colours for each option, none of them are `tfstyleguide` vars.

PR #444, opened 2019-11-19
use event management to close the concierge, not a curtain

alis0nc:


src/AppBar/ConciergeModal.jsx
```diff
@@ -11,60 +12,107 @@ const BetaBadge = () => (
...
+    this.setWrapperRef = this.setWrapperRef.bind(this);
+    this.handleClickOutside = this.handleClickOutside.bind(this);
+  }
+
+  componentDidMount() {
+    document.addEventListener('mousedown', this.handleClickOutside);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('mousedown', this.handleClickOutside);
```
ripleyaffect: I'd recommend using `mouseup` for these, to mimic the behavior of a button

src/AppBar/ConciergeModal.jsx
```diff
@@ -11,60 +12,107 @@ const BetaBadge = () => (
...
+   * Set the wrapper ref
+   */
+  setWrapperRef(node) {
+    this.wrapperRef = node;
+  }
+
+  /**
+   * Alert if clicked on outside of element
+   */
+  handleClickOutside(event) {
```
ripleyaffect: I think `handleClick` or `handleMouseup` is clearer, since you're checking if the click was actually outside in this method

src/AppBar/ConciergeModal.jsx
```diff
@@ -11,60 +12,107 @@ const BetaBadge = () => (
...
+      && !this.wrapperRef.contains(event.target)
+      && !eventComesFromOpenCloseButton) {
+      toggleConcierge();
+    }
+  }
+
+  render() {
+    const { toggleConcierge } = this.props;
+    return (
+      <div ref={this.setWrapperRef}>
```
ripleyaffect: the [`createRef` pattern](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) is preferred

src/AppBar/ConciergeModal.jsx
```diff
@@ -11,60 +12,107 @@ const BetaBadge = () => (
...
+    this.wrapperRef = node;
+  }
+
+  /**
+   * Alert if clicked on outside of element
+   */
+  handleClickOutside(event) {
+    const { visible, toggleConcierge } = this.props;
+    const eventComesFromOpenCloseButton =
+      get(event, 'target.parentElement.className', '').indexOf(FILTER_CLASS) >= 0
```
seanohue: This could use lodash#includes to be more semantic and null-safe

src/AppBar/ConciergeModal.jsx
```diff
@@ -11,60 +12,107 @@ const BetaBadge = () => (
...
+      || get(event, 'target.className', '').indexOf(FILTER_CLASS) >= 0;
+    if (visible
+      && this.wrapperRef
+      && !this.wrapperRef.contains(event.target)
+      && !eventComesFromOpenCloseButton) {
+      toggleConcierge();
```
ripleyaffect: the indentation here doesn't delimit the inside of the conditional from the condition should be:
```es6
if (
   visible &&
   this.wrapperRef &&
   !this.wrapperRef.contains(event.target) &&
   !eventComesFromOpenCloseButton
) {
  toggleConcierge();
}
```

src/AppBar/ConciergeModal.jsx
```diff
@@ -11,60 +12,107 @@ const BetaBadge = () => (
...
+class ConciergeModal extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.setWrapperRef = this.setWrapperRef.bind(this);
+    this.handleClickOutside = this.handleClickOutside.bind(this);
+  }
+
+  componentDidMount() {
+    document.addEventListener('mousedown', this.handleClickOutside);
```
ripleyaffect: I think putting this on the `window` is preferred? probably not a huge deal, but typically we use `global` for these. In the webpack build process, `global` is turned into the appropriate variable for the target of the build. In this case, the target is the browser, so `global` becomes `window`.

src/AppBar/ConciergeModal.jsx
alis0nc: `className` is a space-delimited string though, not a collection (yeah I thought this was weird too)

src/AppBar/ConciergeModal.jsx
alis0nc: I like `handleMouseup` because it reflects the name of the event we're capturing. also I got rid of the (inaccurate) comment above this function

src/AppBar/ConciergeModal.jsx
alis0nc: (well, `className` is a `SVGAnimatedString` if we click on a path within a SVG, but I ignored that because the event seems to fire on the `svg` as well.)

src/AppBar/ConciergeModal.jsx
alis0nc: `window` doesn't seem to work, but `global` does

PR #447, opened 2019-11-21
Add tooltip that calls out concierge

alis0nc:
Point students to new concierge by a tooltip. It doesn't disappear until click the concierge (uses cookie).

Screen Shot 2019-11-21 at 14 30 20

### DIFF
--- a/less/modules/app-bar.less
+++ b/less/modules/app-bar.less
@@ -101,7 +101,7 @@
   overflow: hidden;
   opacity: 0;
   transform: translate(-100%, @nav-height);
-  z-index: 0;
+  z-index: 5;
 }
 
 .tui-app-bar-modal__visible {
@@ -165,7 +165,6 @@
 
 .tui-concierge-tooltip {
   .tui-app-bar-modal;
-  .tui-app-bar-modal__visible;
   padding: 6px 8px;
   background-color: @purple;
   height: 49px;
@@ -177,6 +176,7 @@
   display: flex;
   flex-direction: row-reverse;
   align-items: center;
+  z-index: 3;
 
   h2 {
     font-size: 16px;
@@ -201,6 +201,10 @@
   }
 }
 
+.tui-app-nav__concierge-tooltip-visible .tui-concierge-tooltip {
+  .tui-app-bar-modal__visible;
+}
+
 .tui-concierge-modal {
   .tui-app-bar-modal;
   padding: 42px 15px;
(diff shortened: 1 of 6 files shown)


less/modules/app-bar.less
```diff
@@ -230,10 +273,27 @@
   }
 }
 
+.tui-concierge-arrow {
+  .modal-callout;
+  border-bottom-color: @white;
+  z-index: 6; // this is +1 than the concierge's because we want the arrow
```
seanohue: If the z-index for the concierge is pulled into a var, then imported here, you could do `@concierge-z-index + 1`.

Not that I have anything against code comments.

seanohue: Originally I was thinking that making the concierge z-index into a variable and then adding one to it could be good, but the comment here is still helpful since it explains _why_ we're adding one. 🆒

src/AppBar/AppBar.jsx
```diff
@@ -25,10 +29,14 @@ class AppBar extends React.Component {
     this._handleMouseEnter = this._handleMouseEnter.bind(this);
     this._handleMouseLeave = this._handleMouseLeave.bind(this);
     this._shouldInitNotifications = this._shouldInitNotifications.bind(this);
+    this._shouldShowTooltip = this._shouldShowTooltip.bind(this);
+    this._setTooltipDismissed = this._setTooltipDismissed.bind(this);
+    this.cookies = new Cookies();
```
seanohue: 🍪

src/AppBar/AppBar.jsx
```diff
@@ -84,9 +96,35 @@ class AppBar extends React.Component {
...
+    return !this.cookies.get(TOOLTIP_KEY);
+  }
+
+  _setTooltipDismissed() {
+    this.cookies.set(
+      TOOLTIP_KEY,
+      true,
+      {
+        path: '/',
+        expires: moment().add(1, 'month').toDate(),
```
seanohue: Does this mean we show them the tooltip approx. once per month?

alis0nc: Yep, this is subject to change if I get any special requests from Ellen

src/AppBar/AppBar.jsx
alis0nc: yummy yummy

src/AppBar/AppBar.jsx
alis0nc: Ellen says this is ok

PR #453, opened 2019-12-17
T12713 fix domain on concierge cookie

elizabethwyatt:
When switching between subdomains, the concierge cookie isn't being shared, so a student going from their dashboard to their curriculum might see the concierge tooltip twice.

This update will replace the old cookie with one that's accessible across subdomains.


src/AppBar/AppBar.jsx
```diff
@@ -127,19 +128,24 @@ class AppBar extends React.Component {
     return !/mentor|admin/.test(user.role) && user.platform === LEGACY_PLATFORM;
   }
 
-  _shouldShowTooltip() {
-    return !this.cookies.get(TOOLTIP_KEY);
-  }
-
   _setTooltipDismissed() {
     this.cookies.set(TOOLTIP_KEY, true, {
+      domain: location.hostname == 't.ful' ? 't.ful' : '.thinkful.com',
```
alis0nc: nice

PR #457, opened 2020-02-13
T13311 Add concierge prep feature flag and check for prep enrollment.

seanohue:
Bump version, add new feature flag to concierge as well as check for prep enrollment

This is to enable testing cards such as https://phabricator.bloc.io/T13311 and the concierge stories that follow it.


src/AppBar/AppBar.jsx
```diff
@@ -42,11 +43,12 @@ class AppBar extends React.Component {
     this.cookies = new Cookies();
 
     this.state = {
-      isMenuVisible: false,
       isConciergeVisible: false,
       isConciergeTooltipVisible: true,
-      slackUrl: null,
+      isMenuVisible: false,
```
seanohue: Had to alphabetize these 😈

PR #458, opened 2020-02-18
Messed up the release of T13311 -- this gets the flag right.

seanohue:
Missed a spot where the flag is, and also should bind method to the component...


src/AppBar/AppBar.jsx
```diff
@@ -222,7 +223,7 @@ class AppBar extends React.Component {
             onMouseEnter={this._handleMouseEnter}
             EnrollmentView={EnrollmentView}
           />
-          {_.includes(user.access, CONCIERGE_FLAG) && (
+          {this._hasConciergeAccess(user) && (
```
learykara: no need to change but fwiw since `user` is in `this.props`, you could pull it directly in `_hasConciergeAccess` instead of passing it as a prop

PR #463, opened 2020-02-25
Use archives link to direct students to correct namespace.

seanohue:
https://phabricator.bloc.io/T13538

Without this change, if a student is not signed into the Thinkful slack yet they get redirected to a generic slack login. With this change they can sign up for the Thinkful slack more easily.


seanohue:
Closing since this got solved in Pysplash instead

PR #475, opened 2020-08-12
[TFOP-186] update dependencies, fix lodash usage

thasmin:
- I had to manually change the tf-analytics dependency to be `file:../tf-analytics` to get it to work. I don't know how to deploy that project.
- The major version bump to qs was only to remove backwards compatibility with versions we don't use.
- Lodash required some minor changes to the AvailabilityGrid but it still works in Storybook:
image


thasmin:
We should merge and deploy https://github.com/Thinkful/tf-analytics/pull/12 first.

seanohue:
> We should merge and deploy [Thinkful/tf-analytics#12](https://github.com/Thinkful/tf-analytics/pull/12) first.

That makes sense. To be honest I am not sure how to deploy it.

PR #476, opened 2020-08-19
[TFOP-186] update more dependencies

thasmin:
https://chegg.atlassian.net/browse/TFOP-186

This step required upgrading storybook and the whole webpack chain.
Upgrading storybook required changing to the new config scheme in main.js. The actions addon was unused and removed.
Upgrading babel required removing stage0 in favor of the one piece we used, plugin-proposal-class-properties.

All components in storybook worked without changes. The one blemish is that the image in the AppBar does not load because Storybook doesn't send an authorization header. When I copy and paste the image's URL, I get the Zaphod pic as expected.
image


.storybook/main.js
```diff
@@ -0,0 +1,23 @@
...
+        'less-loader',
+      ],
+      include: [
+        path.resolve(__dirname, '../less'),
+        path.resolve(__dirname, '../node_modules/tfstyleguide'),
+      ]
+    });
+    return config;
+  },
+};
```
seanohue, line 23: *nit:* newline EOF

seanohue:
When testing this locally and trying to view storybook, `./script/server` results in the following error:
```
/Users/seanodonohue/Thinkful/ui/node_modules/fs-extra/lib/mkdirs/make-dir.js:85
      } catch {
```

Unless I use a newer Node version than some of our projects. We could update the `package.json` to specify this, or use an `.nvmrc`

PR #478, opened 2020-09-01
[TFOP-227] update react, add jest/enzyme/simple test

thasmin:


thasmin:
Good thought. I'll add the circle check to this PR. I think it goes better here.

